PR #213, opened 2022-02-09
fix: Refactor the "basic" component to use nodeOperators

abensonca:
Cosmic time and DMO mass evolution are now handled by nodeOperators.

The "unevolving" basic component is removed - a new nodeOperator allows for non-evolution of the DMO mass.

The "massTarget" property of the basic component is replaced by a meta-property added by the relevant nodeOperator.


abensonca:
This change to Galacticus will require you to add in two extra "`nodeOperators`" to parameter files. If you already have a section like this:
```
  <!-- Node evolution and physics -->
  <nodeOperator value="multi">
    <nodeOperator value="....."/>
    <nodeOperator value="....."/>
    .
    .
    .
  </nodeOperator>
```
then you can simply add these new operators at the start of this section, like this:
```
  <!-- Node evolution and physics -->
  <nodeOperator value="multi">
    <!-- Cosmological epoch -->
    <nodeOperator value="cosmicTime"/>
    <!-- DMO evolution -->
    <nodeOperator value="DMOInterpolate"/>
    <!-- Other stuff -->
    .
    .
    .
  </nodeOperator>
```
The first of these new operators, "`cosmicTime`", just evolves the cosmic time forward in each node. The second, "`DMOInterpolate`", linear interpolates the dark-matter-only (DMO) mass of each node between parent and child halos. Previously these actions were taken automatically by the "`basic`" component of each node.

PR #257, opened 2022-07-13
Use the k most bound particles to determine the center of a halo when doing self-bound analysis

Xiaolong-Du:
When doing self-bound analysis on N-body simulation data, now it allows to use the k most bound particles to determine the center of a halo.

Changes made:
1. The updates in master branch are first merged into the orbitCalibration branch;
2. An interface to the the GSL sorting function that finds the k smallest or largest elements is added;
3. Allow to set the number of representative particles by the parameters "representativeFraction" and "representativeMinimumCount".


abensonca:
Looks good to merge into the `orbitCalibration` branch. 

Do you also plan to merge this into `master`?

The four failing tests all seem to be caused by the website for the ANN library being down right now - so are not a problem.

Xiaolong-Du:
> Looks good to merge into the `orbitCalibration` branch.
> 
> Do you also plan to merge this into `master`?
> 
> The four failing tests all seem to be caused by the website for the ANN library being down right now - so are not a problem.

Yes, it will be good to merge it also into the master branch. The orbitCalibration branch contain a few useful things, e.g. using tree method to compute the gravitational potential between N-body particles, speeding up the calculation of generalized hypergeometric function, etc.

PR #258, opened 2022-07-13
Fix issues with the fuzzy dark matter transfer function

Xiaolong-Du:
1. In some of the fuzzy dark matter transfer function classes, the quarter-mode mass is not defined;
2. In "Murgia2017"  transfer function class, the definition of half-mode mass is different from the one used at other places. There are two definitions used in literature:
    a) T^2(k_half) = 1/2;
    b) T(k_half) = 1/2.
In Galacticus, we use the latter.


Xiaolong-Du:
> This looks good and is ok to merge into master.
> 
> The definition of half-mode mass as T(khalf)=1/2 is a good catch! It would be good to add a short section on this in the `doc/Definitions.tex` file for future reference.

Good idea. I will add the description of half-mode mass to the documentation.

PR #267, opened 2022-07-27
Ensure axis ratios are ordered in the `inertiaTensor` N-body operator

abensonca:


Xiaolong-Du:
Due to the changes in commit 59551cc, `indexMostBound` now stores the indices of the k most bound particles, i.e. it becomes a 2D array. Some minor changes are needed to make sure the calculations in the `inertiaTensor` N-body operator works properly.

Xiaolong-Du:
> Thanks @Xiaolong-Du - do you have a patch for this? If not, is the correct solution in this case to just make `indexMostBound` a 2-D array and use the first entry (i.e. the most bound particle) in computing the inertia tensor?

I do not have a patch for this. But the solution you mentioned should solve the problem. As you said, the first entry of `indexMostBound` is the index of the most bound particle. We can also make use of all the k most bound particles to estimate the halo center. We can always set k=1 in the parameter file so that it uses only the most bound particle.

abensonca:
OK - sounds good. I'll make that change. Thanks!

Xiaolong-Du:
No problem.

By the way, the parameters that control `k` are `representativeMinimumCount` and `representativeFraction`. The default values of these two parameters are 10 and 0.05, respectively. To have `k=1`, one just needs to set `representativeMinimumCount=1 `and `representativeFraction=0`.

PR #274, opened 2022-08-18
Implement a hyperbolic tangent power spectrum window function.

Xiaolong-Du:
Compared to the ETHOS or smooth-k filter, the hyperbolic tangent filter can deal with a power spectrum that is more sharply suppressed on small scales. When the smoothing scale is below the half-mode scale, this new filter can mitigate the artificial contribution to the cosmic variance from large-scale power.


abensonca:
Looks good - thanks @Xiaolong-Du !

PR #303, opened 2022-09-26
fix: Fix bug in array offsets used for property scales during ODE solution

abensonca:
Was prevoiusly using the "all" offsets, but should be using "active" offsets since only active properties are evolved. This resulted in incorrect scales being applied for some properties.


abensonca:
Invalid solution - closing.

PR #306, opened 2022-10-06
Prevent the `mergerTreeEvolveTimestepHostTidalMassLoss` from taking tiny timesteps

abensonca:
If tree evolution is using the `backtrackToSatellites` option, this timestep limiter could lead to extremely tiny timesteps for a satellite as it is allowed to evolve only a slight amount beyond the time of its host. This fix prevents this by refusing the evolve the satellite (if this class is the limiting factor in evolution) unless the step that can be taken is a sufficiently large fraction of the mass loss timescale. If a step is refused, this will force the host halo to evolve instead. This evoids tiny timesteps.


Xiaolong-Du:
@abensonca That is an interesting issue. Your solution looks reasonable to me. I will take a look at the changes later today.

Xiaolong-Du (commented):
The changes look good to me.

PR #338, opened 2022-12-02
Add functionality for building constrained merger trees

abensonca:
Implements a Brownian bridge-based approach to constructing merger trees which are guaranteed to have a branch passing through a specific target (mass,redshift).


abensonca:
@Xiaolong-Du The constrained merger tree code is almost ready to merge. There's just one test that fails:

[`cole2000TreeBuilderWDMNodeWellOrdering.xml`](https://github.com/galacticusorg/galacticus/blob/constrainedTrees/testSuite/regressions/cole2000TreeBuilderWDMNodeWellOrdering.xml) 

It uses the `farahiMidpoint` solver for a WDM model. It seems to be working but gets cancelled because it takes too long. This seems to be because of the fix that makes the non-crossing rates be retabulated every time that the mass corresponding to the maximum variance changes.

I don't think there's any way around this slow-down, but wanted to get your input on this. If not, I'll see if I can speed up the test by lowering accuracy and/or resolution.

Xiaolong-Du:
Hi @abensonca, I will try if I can find a solution. But it is a bit strange that the mass corresponding to the maximum variance will change frequently. When `farahiRateNonCrossingInterpolate` is called, the maximum variance will be set to the variance corresponding to `massMinimum` specified in the `mergerTreeBranchingProbabilityGnrlzdPrssSchchtr` class.

abensonca:
Doesn't the mass corresponding to `varianceMaxinum` change as a function of time?

Xiaolong-Du:
It should not. But since we do conversions from mass to variance and then back to mass, there could be some numerical errors. So when checking whether the mass corresponding to the maximum variance changes, I added a tolerance. It may help if we allow a larger tolerance, i.e. a larger value for `toleranceRelativeMass` in the function `farahiRateNonCrossingInterpolate`.

abensonca:
Ok, that makes sense. I'll try adjusting the tolerance. Thanks!

Xiaolong-Du:
Hi @abensonca , I ran some tests. I think the main issue is with the crossing rate tabulation. There was a function `excursionSetTest` in the `mergerTreeBranchingProbabilityGnrlzdPrssSchchtr` class previously. It makes sure that the excursion set solver tabulates crossing rates for a sufficiently large mass range when it is first called. This function is deleted in the `constrainedTrees` branch for some reason. So Galacticus needs to do retabulations each time the crossing rate is requested for a variance 2 times larger than the previous call. To solve this issue, we can either add back the `excursionSetTest` function or make the table extendable also in the variance direction.

Xiaolong-Du:
Besides, the parameter `timeStepFractional` is renamed as `fractionalTimeStep` in the `constrainedTrees` branch. So we will also need to make the change in the parameter file.

abensonca:
> Hi @abensonca , I ran some tests. I think the main issue is with the crossing rate tabulation. There was a function `excursionSetTest` in the `mergerTreeBranchingProbabilityGnrlzdPrssSchchtr` class previously. It makes sure that the excursion set solver tabulates crossing rates for a sufficiently large mass range when it is first called. This function is deleted in the `constrainedTrees` branch for some reason. So Galacticus needs to do retabulations each time the crossing rate is requested for a variance 2 times larger than the previous call. To solve this issue, we can either add back the `excursionSetTest` function or make the table extendable also in the variance direction.

This function was removed in 00ea0e23e3f6907aabc6a494450bf30b94322520 because it seemed less useful now that the variance at fixed mass grows with time. But maybe it is actually beneficial still. I'll try re-adding it.

abensonca:
> Besides, the parameter `timeStepFractional` is renamed as `fractionalTimeStep` in the `constrainedTrees` branch. So we will also need to make the change in the parameter file.

Thanks - I've updated these parameter files.

Xiaolong-Du:
> It seems that in some models (e.g. WDM) the numerical errors can be very large, because σ(_M_) becomes a very weak function of _M_ at low masses.
> 
> But, the `cosmologicalMassVariance` class has a function `growthIsMassDependent()` which is true if the growth factor is a function of mass, and false otherwise. So, if this function is false I think we should never need to retabulate the non-crossing rate. Is that correct?
> 
> If so, I can change the logic to check this function first, and then only retabulate if this function is true _and_ the minimum mass changes.

I think a better way would be using the minimum mass as an input to the the non-crossing rate function instead of the maximum variance. Then we should have no problem no matter `cosmologicalMassVariance` has a mass dependent growth or not. At first, I use the maximum variance as an input argument so that the analytic non-crossing rate function for a linear barrier is a function of S and S_max. But that is not really necessary. As you said, for some special cases, the numerical errors when converting  σ(_M_) to M can be very large, it would be better to use the minimum mass directly.

Xiaolong-Du:
A potential issue with commit [eef87da](https://github.com/galacticusorg/galacticus/pull/338/commits/eef87da1d3cd4e27c55e4e012696d2aaa88a8fcc) is that when the non-crossing rate table is read in, `tableInitializedNonCrossingRate` is not set to `.true.`, so Galacticus will retabulate the non-crossing rate even if we already have the solution read from the file.

Xiaolong-Du:
> > I think a better way would be using the minimum mass as an input to the the non-crossing rate function instead of the maximum variance. Then we should have no problem no matter `cosmologicalMassVariance` has a mass dependent growth or not. At first, I use the maximum variance as an input argument so that the analytic non-crossing rate function for a linear barrier is a function of S and S_max. But that is not really necessary. As you said, for some special cases, the numerical errors when converting σ(_M_) to M can be very large, it would be better to use the minimum mass directly.
> 
> That sounds like a good idea. Can you take a look at [b2849ba](https://github.com/galacticusorg/galacticus/pull/338/commits/b2849bac24f7d51a571147a5a85e61fbd69b979d) where I've attempted to implement this and see if it looks reasonable to you?

Yes, I have looked through the modifications. They look fine to me.

abensonca:
Thanks! I'll wait and see if this now passes all of the tests - if it does I think it's ready to merge.

PR #357, opened 2023-01-18
Fix an issue that the progress bar is inaccurate when solving the excursion sets problem with MPI

Xiaolong-Du:
The progress bar is inaccurate when we use multiple MPI processes and `self%coordinatedMPI_` is `.true.`.


abensonca (approved):
This all looks good - thanks @Xiaolong-Du !

PR #371, opened 2023-02-07
Allow σ(M) integral to be split at local minima of the transfer function

abensonca:
This makes integration more robust for oscillating transfer functions for example.


abensonca:
For example - this shows the transfer function for a fuzzy dark matter model, which has very complicated structure at high wavenumber. This patch automatically identifies the local minima and splits the $\sigma(M)$ integral at those points.

tk

PR #411, opened 2023-05-24
Fix bugs in the implementation of pre-infall evolution

Xiaolong-Du:
When doing pre-infall evolution of a halo, accounting for the orbit evolution, tidal mass loss and heating, one needs to find the host halo at the corresponding time (before infall). In Galacticus, it walks up a branch in the merge tree to find the host node:

```
    nodeHost => nodeProgenitor
    do while (nodeHost%isPrimaryProgenitor())
       nodeHost => nodeHost%parent
    end do
    basicProgenitor => nodeHost%basic()
    do while (associated(nodeHost) .and. basicProgenitor%time() > time)
       nodeHost => nodeHost%firstChild
       if (associated(nodeHost)) basicProgenitor => nodeHost%basic()
    end do
```

There is a bug in the above code. For example, if we have a merger tree like the one in the attached plot. For `nodeProgenitor => node6`, the code give `nodeHost => node6`, this is incorrect. The host of `node6` at time `t` should be `node5`. On the other hand, for `node5`, which is the primary progenitor of `node4`, the host at `t` should be `node3`.

Modifying the above code as follows solves the issue:

```
    nodeHost => nodeProgenitor
    do while (nodeHost%isPrimaryProgenitor())
       nodeHost => nodeHost%parent
    end do
    nodeHost        => nodeHost%parent%firstChild
    basicProgenitor => nodeHost%basic()
    do while (associated(nodeHost) .and. basicProgenitor%time() > time)
       nodeHost => nodeHost%firstChild
       if (associated(nodeHost)) basicProgenitor => nodeHost%basic()
    end do
```

tree 001

### DIFF
--- a/source/nodes.operators.physics.satellite.heating.tidal.F90
+++ b/source/nodes.operators.physics.satellite.heating.tidal.F90
@@ -148,6 +148,7 @@ contains
        do while (nodeHost%isPrimaryProgenitor())
           nodeHost => nodeHost%parent
        end do
+       nodeHost  => nodeHost%parent%firstChild
        ! Follow that node back through descendants until we find the node at the corresponding time.
        basic     => node    %basic()
        basicHost => nodeHost%basic()
--- a/source/nodes.operators.physics.satellite_orbits.F90
+++ b/source/nodes.operators.physics.satellite_orbits.F90
@@ -171,7 +171,7 @@ contains
     time               =  basicHost     %time       (                 )
     ! Integrate the orbit backward in time to each progenitor halo.
     allocate(solver)
-    solver         =  odeSolver(6_c_size_t,orbitalODEs,toleranceRelative=1.0d-3)
+    solver         =  odeSolver(6_c_size_t,orbitalODEs,toleranceRelative=1.0d-3,toleranceAbsolute=1.0d-6)
     self_          => self
     do while (associated(nodeProgenitor))
        basicProgenitor     => nodeProgenitor %basic             (                 )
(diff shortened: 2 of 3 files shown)


Xiaolong-Du:
There exists another bug in the orbit evolution module. If `trackPreInfallOrbit` is switched on, Galacticus will track the orbit evolution of halos before they fall into the host's viral radius. This is done by first evolving the satellite orbit backward from its infall position `x0` (with infall velocity `v0`) to the position of its progenitor halo `x1` in the host potential. Then an effective velocity is computed as `veff=(x0-x1)/(t0-t1)`. When the halo is evolved forward, it is assigned a constant velocity `veff` (neglecting the host potential) so that the halo will reach the position `x0` when it enters the host's virial radius. However, the velocity of satellite at infall is then `veff`, which does not match the infall velocity `v0`. So we need to reset the velocity of satellites at infall. This is fixed in the commit f79d006c825b332cb13ea055138ffa1a2a92ba34.

An example is shown in the attached plot. The satellite fall into the host's virial radius at `t=6Gyr`. The blue curve shows the orbit evolution of the satellite from the current version of Galacticus. As we can see, the orbit of satellite is changed after infall, i.e. if we now evolve the satellite's orbit backward from infall in the host potential, it will not reach `R~0.6` Mac at `t=1Gyr`. The orange curve shows the result after the above modification. If we shift the orange curve to the left, we can see that it reaches `R~0.6` Mac at `t=1Gyr` as expected (green dashed curve). Note that the orange curve and green curve do not match at `1Gyr < t < 6 Gyr`. This is expected because we assume a constant effective velocity when evolving the satellite forward in time from `1Gyr` to `6Gyr`.

R_orbit_pre_infall

abensonca (approved):
These changes all look good.

abensonca:
Everything looks good - will merge this PR now.

PR #412, opened 2023-05-24
Fix bugs in the calculations of tidal radius

Xiaolong-Du:
When calculating the tidal radius, Galacticus will compute the eigenvectors of the tidal tensor and project the tidal acceleration in the direction of the eigenvector with the largest eigenvalue.

```
       tidalFieldRadial                =-abs(tidalTensor%vectorProject(tidalTensorEigenVectorComponents(maxloc(tidalTensorEigenValueComponents,dim=1),:))) &
            &                           *(                                                                                                                 &
            &                             +kilo                                                                                                            &
            &                             *gigaYear                                                                                                        &
            &                             /megaParsec                                                                                                      &
            &                            )**2
```

There are three issues:
1) Different eigenvectors are stored in different columns of the matrix `tidalTensorEigenVectorComponents`. See the documentation of the function [`GSL_Eigen_SymmV`](https://www.gnu.org/software/gsl/doc/html/eigen.html). So we should take the column that corresponding to the largest eigenvalue, not the row unless the matrix `tidalTensorEigenVectorComponents` is symmetric (usually it is not).

2) In some special cases, the tidal force can be compressive. In such cases, the projection of the tidal tensor becomes negative. If we take the absolute value and multiply it by `-1` as shown above, `tidalFieldRadial` is always negative. So even when the tidal force is compressive, we will get a non-zero tidal mass loss rate.

3) The projection of the tidal tensor in the direction of a eigenvector is exactly the eigenvalue corresponding to that eigenvector, so we can simply use the eigenvalue instead of computing the projection.

### DIFF
--- a/source/satellites.tidal_stripping.radius.King1962.F90
+++ b/source/satellites.tidal_stripping.radius.King1962.F90
@@ -290,8 +290,9 @@ contains
        tidalFieldRadial                =+0.0d0
     end if
     ! If the tidal force is stretching (not compressing), compute the tidal radius.
+    tidalPull=frequencyAngular**2-tidalFieldRadial
     if     (                                                                          &
-         &   frequencyAngular**2                                   > tidalFieldRadial &
+         &   tidalPull                                             >  0.0d0           &
          &  .and.                                                                     &
          &   massSatellite                                         >  0.0d0           &
          &  .and.                                                                     &
@@ -300,7 +301,6 @@ contains
        ! Check if node differs from previous one for which we performed calculations.
        if (node%uniqueID() /= self%lastUniqueID) call self%calculationReset(node)
        ! Initial estimate of the tidal radius.
-       tidalPull=frequencyAngular**2-tidalFieldRadial
        if (self%radiusTidalPrevious <= 0.0d0) then
           self%radiusTidalPrevious=+sqrt(                                              &
                &                         +gravitationalConstantGalacticus              &
--- a/source/satellites.tidal_stripping.radius.King1962.F90
+++ b/source/satellites.tidal_stripping.radius.King1962.F90
@@ -280,7 +280,7 @@ contains
        call tidalTensorMatrix%eigenSystem(tidalTensorEigenVectors,tidalTensorEigenValues)
        tidalTensorEigenValueComponents = tidalTensorEigenValues
        tidalTensorEigenVectorComponents= tidalTensorEigenVectors
-       tidalFieldRadial                =-abs(tidalTensor%vectorProject(tidalTensorEigenVectorComponents(maxloc(tidalTensorEigenValueComponents,dim=1),:))) &
+       tidalFieldRadial                =-abs(tidalTensor%vectorProject(tidalTensorEigenVectorComponents(:,maxloc(tidalTensorEigenValueComponents,dim=1)))) &
             &                           *(                                                                                                                 &
             &                             +kilo                                                                                                            &
             &                             *gigaYear                                                                                                        &
--- a/source/satellites.tidal_stripping.radius.King1962.F90
+++ b/source/satellites.tidal_stripping.radius.King1962.F90
@@ -280,11 +280,11 @@ contains
        call tidalTensorMatrix%eigenSystem(tidalTensorEigenVectors,tidalTensorEigenValues)
        tidalTensorEigenValueComponents = tidalTensorEigenValues
        tidalTensorEigenVectorComponents= tidalTensorEigenVectors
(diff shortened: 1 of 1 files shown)


source/satellites.tidal_stripping.radius.King1962.F90
```diff
@@ -279,12 +280,11 @@ double precision function king1962Radius(self,node)
        tidalTensorMatrix               = tidalTensorComponents
        call tidalTensorMatrix%eigenSystem(tidalTensorEigenVectors,tidalTensorEigenValues)
        tidalTensorEigenValueComponents = tidalTensorEigenValues
-       tidalTensorEigenVectorComponents= tidalTensorEigenVectors
-       tidalFieldRadial                =-abs(tidalTensor%vectorProject(tidalTensorEigenVectorComponents(maxloc(tidalTensorEigenValueComponents,dim=1),:))) &
-            &                           *(                                                                                                                 &
-            &                             +kilo                                                                                                            &
-            &                             *gigaYear                                                                                                        &
-            &                             /megaParsec                                                                                                      &
+       tidalFieldRadial                =-maxval(tidalTensorEigenValueComponents) &
```
abensonca, line 283: Maybe we should add an extra check here to ensure that `maxval(tidalTensorEigenValueComponents)` is always positive - to avoid stripping for compressive tidal fields? In practice, I think it's physically very unlikely (in the types of system we consider) to have a tidal field that is compressive in all directions - but I think it is technically possible.

So, we could change this to:
```
       tidalFieldRadial                =-max(maxval(tidalTensorEigenValueComponents),0.0d0) &
            &                           *(                                                  &
            &                             +kilo                                             &
            &                             *gigaYear                                         &
            &                             /megaParsec                                       &
            &                            )**2
```

abensonca (changes requested):
This looks good overall. I left one comment - I think we should catch cases where all eigenvalues are negative (i.e. a tidal field compressive in all dimensions) and simply truncate to zero.

abensonca:
I checked the effects of this fix on the subhalo population statistics. The following plots show the standard validation tests that Galacticus makes for a dark matter-only subhalo model. The left column is before this fix, the right column is after this fix.
validation
There are no significant differences, so I think that this bug didn't strongly affect the results at the population level.

source/satellites.tidal_stripping.radius.King1962.F90
Xiaolong-Du: Hi Andrew, yes, I think you are right. I did not do that at first because there is a check later

```
    ! If the tidal force is stretching (not compressing), compute the tidal radius.
    if     (                                                                          &
         &   frequencyAngular**2                                   > tidalFieldRadial &
         &  .and.                                                                     &
         &   massSatellite                                         >  0.0d0           &
         &  .and.                                                                     &
         &   self%galacticStructure_%massEnclosed(node,radiusZero) >= 0.0d0           &
         & ) then
```

Here it check whether the centrifugal force+tidal force is compressive. But actually if there is no tidal force, we should expect zero tidal mass loss. So besides the change you suggested, we might also need to change the above lines to

```
    ! If the tidal force is stretching (not compressing), compute the tidal radius.
    if     (                                                                          &
         &   tidalFieldRadial                                      >  0.0d0           &
         &  .and.                                                                     &
         &   massSatellite                                         >  0.0d0           &
         &  .and.                                                                     &
         &   self%galacticStructure_%massEnclosed(node,radiusZero) >= 0.0d0           &
         & ) then
```

abensonca: Good point. But, even if there is no tidal field, in principle the centrifugal force can lead to some stripping I think? If the centrifugal force in the outer parts of the subhalo exceeds the self-gravity of the subhalo then particles could be lost. If that's the case then I think it would be best to keep:
```
       tidalFieldRadial                =-maxval(tidalTensorEigenValueComponents) &
            &                           *(                                        &
            &                             +kilo                                  &
            &                             *gigaYear                              &
            &                             /megaParsec                            &
            &                            )**2
```
and then keep the original check for compressive forces, but maybe for simplicity and clarity change it to:
```
    ! If the tidal force is stretching (not compressing), compute the tidal radius.
    tidalPull=frequencyAngular**2-tidalFieldRadial
    if     (                                                                          &
         &   tidalPull                                             >  0.0d0           &
         &  .and.                                                                     &
         &   massSatellite                                         >  0.0d0           &
         &  .and.                                                                     &
         &   self%galacticStructure_%massEnclosed(node,radiusZero) >= 0.0d0           &
         & ) then
(and remove the subsequent calculation of  `tidalPull` on the next line.

Does that make sense?

Xiaolong-Du: Yes, that makes sense. I will make the change as you suggested.

abensonca (approved):
Thanks - looks good now!

abensonca:
Merging this fix now.

PR #413, opened 2023-05-24
Track peak bound masses of nodes

abensonca:
Adds a `nodeOperator` and a `nodePropertyExtractor` to allow tracking and output of the peak bound mass ever attained by each halo.


abensonca:
@Xiaolong-Du These should allow you to output the peak mass of each halo. You'll need to add these parameters:
```
<nodeOperator value="massBoundMaximum"/>

<nodePropertyExtractor value="massBoundMaximum" />
```
The results from a test model look reasonable, but let me know if you see any problems or unexpected behavior.

Xiaolong-Du:
Thanks! I will run some tests.

Xiaolong-Du:
Hi @abensonca , I have run several tests. The results look fine.

abensonca:
Thanks! In that case I'll go ahead and merge this PR.

PR #417, opened 2023-06-01
Improve the calculation of radial velocity dispersion for heated dark matter profile

Xiaolong-Du:
For heated dark matter profiles, the radius velocity dispersion is computed by numerically integrating the Jeans equation:

$\sigma_r(r) = \frac{1}{\rho(r)}\int_r^{r^{\mathrm{max}}} \rho(r) \frac{\mathrm{G} M(r)}{r^2} \mathrm{d} r.$

This is time consuming because one needs to solve for $\rho(r)$ frequently. If we do the above integration with respect to the initial radius $r_i$ which is related to $r$ by

$-\frac{G M}{2 r_i} + \Delta E_h(r_i) = -\frac{G M}{2 r},$

it will be much faster, i.e.

$\sigma_r(r) =\frac{1}{\rho(r)}\int_{r_i}^{r_{i}^{\mathrm{max}}} \rho_i(r_i) \frac{\mathrm{G} M(r_i)}{r_i^2}\left(\frac{r_i}{r}\right)^4 \mathrm{d} r_i$.

Here $\Delta E_h$ is the heating energy on a mass shell with initial radius $r_i$. Then one only needs to compute $\rho(r)$ once and $\rho_i(r_i)$ is usually very easy to calculate.

Computing time for running a merger tree with $M_{\rm host}=10^{13}M_{\odot}$ and $M_{\rm res}=2.4 \times 10^9 M_{\odot}$:
a) Orignial: 1154.31 seconds;
b) New: 137.74 seconds.

### DIFF
--- a/source/dark_matter_profiles.generic.F90
+++ b/source/dark_matter_profiles.generic.F90
@@ -88,6 +88,8 @@ module Dark_Matter_Profiles_Generic
        <method description="Set a sub-module scope pointers on a stack to allow recursive calls to functions." method="solverSet"/>
        <method description="Unset the sub-module scope pointer." method="solverUnset"/>
        <method description="Reset generic profile memoized calculations." method="calculationResetGeneric"/>
+       <method description="Integrand for dark matter profile Jeans equation." method="jeansEquationIntegrand"/>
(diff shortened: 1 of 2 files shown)


abensonca (approved):
Everything looks good - thanks for figuring out this faster solution!

PR #437, opened 2023-07-20
Fix issues with the calculation of gravitational potential for Zhao1996 profile

Xiaolong-Du:
For certain combination of parameters, e.g. $\alpha=1$, $\beta=3$ and $\gamma=0$, the calculation of the gravitational potential for Zhao1996 density profile requires computing the regularized generalized hypergeometric function **Hypergeometric_pFq_Regularized** at points where the function value is infinite.

The equation of gravitational potential for Zhao1996 density profile can be simplified and written as a function containing the Gaussian hypergeometric function $_2F_1(a_1,a_2;b_1;x)$, which does not have issues in the case mentioned above. For several special cases, we can further simplify the expression and write the potential using only elementary functions.


abensonca (changes requested):
This all looks good - thanks for the fixes and improvements! The new test code should also be added to the [code.yml](https://github.com/galacticusorg/galacticus/blob/master/.github/workflows/cicd.yml) workflow so that it runs automatically on future commits and PRs. You can search in that file for a similar test, e.g. `Tests-Dark-Matter-Profiles`, and copy-paste that and modify as needed. (You should also add the new test to the `needs` list in the `Deploy` job in that file so that this test will be required to pass before the code is accepted and deployed.

Xiaolong-Du:
> This all looks good - thanks for the fixes and improvements! The new test code should also be added to the [code.yml](https://github.com/galacticusorg/galacticus/blob/master/.github/workflows/cicd.yml) workflow so that it runs automatically on future commits and PRs. You can search in that file for a similar test, e.g. `Tests-Dark-Matter-Profiles`, and copy-paste that and modify as needed. (You should also add the new test to the `needs` list in the `Deploy` job in that file so that this test will be required to pass before the code is accepted and deployed.

Got it. I have added the test to the workflow.

abensonca (approved):
Looks good - thanks!

PR #440, opened 2023-07-25
Add a new fitting function of the mass-concentration relation for WDM

Xiaolong-Du:
Implement the fitting function from Bose et al. 2016:

$c_\mathrm{WDM} = c_\mathrm{CDM} \left( 1 + \gamma_1 {M_\mathrm{1/2} \over M_\mathrm{halo}} \right)^{-\gamma_2} (1+z)^{\beta(z)},$

where $\gamma_1=60$, $\gamma_2=0.17$, $M_\mathrm{1/2}$ is the mass corresponding to the wavenumber at which the WDM transfer function is suppressed below the CDM transfer function by a factor of 2, $M_\mathrm{halo}$ is the mass of the dark matter halo, and $\beta(z)=0.026 z-0.04$.

### DIFF
--- a/doc/Galacticus.bib
+++ b/doc/Galacticus.bib
@@ -4844,6 +4844,21 @@ mass-concentration-redshift relation, c(M, z), of dark matter haloes. Our simula
 	file = {NASA/ADS Full Text PDF:/home/abensonca/.mozilla/firefox/f54gqgdx.default/zotero/storage/GA2J7XH3/Ludlow et al. - 2016 - The mass-concentration-redshift relation of cold a.pdf:application/pdf}
 }
 
+@article{bose_copernicus_2016,
+        author = {Bose, Sownak and Hellwing, Wojciech A. and Frenk, Carlos S. and Jenkins, Adrian and Lovell, Mark R. and Helly, John C. and Li, Baojiu},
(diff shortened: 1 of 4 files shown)


Xiaolong-Du:
To compute the halo concentration in WDM models, one can also use the model proposed in [Schneider 2015](https://arxiv.org/abs/1412.2133): find the corresponding CDM halo that forms at the same redshift as the WDM halo and assume that the haloes with the same collapse redshifts end up having the same concentrations. For example, in WDM models, the formation of small halos below the half-mode scale are significantly delayed. So they correspond to CDM halos with much higher masses which also form at late times.

As is found in [Diemer & Joyce 2019](https://arxiv.org/abs/1809.07326), in CDM models, the concentration mass relation have an upturn at high masses. If we use the Schneider 2015 model, the concentration mass relation for WDM will also have an upturn at small masses, see the attached plot. Such phenomenon has been observed in some WDM simulations, e.g. [Schneider 2015](https://arxiv.org/abs/1412.2133). However, as is shown in the attached plot, the halo concentration seems unrealistically high for small WDM halos (orange dashes curve). One solution is to slightly change the original Diemer & Joyce 2019 model.

abensonca (approved):
This all looks good - ok to commit once tests pass. Thanks for the fixes and additions!

PR #454, opened 2023-08-29
A factor of specific reduced mass is missing when propagating the Kepler orbit

Xiaolong-Du:
Given the specific orbital angular momentum $L$ and specific energy $E$ at a radius $r$, the Kepler orbit can be propagated to a new radius $r^{\rm new}$ as:

$V_t=\frac{L}{r^{\rm new} \mu_{\rm reduced}}$,

$V_r=\sqrt{2 \left( E + \frac{G M_{\rm host}}{r^{\rm new}} \right)/\mu_{\rm reduced}-V_t^2}$,

where $V_t$ and $V_r$ are the tangential and radial velocity at the new radius $r^{\rm new}$, respectively. $\mu \equiv 1/(1+M_{\rm sat} / M_{\rm host})$ is the specific reduced mass in the two-body problem.

In the previous implementation, a factor of $\mu_{\rm reduced}$ is missing when computing $V_t$. This is now fixed.

### DIFF
--- a/source/objects.kepler_orbits.F90
+++ b/source/objects.kepler_orbits.F90
@@ -948,7 +948,7 @@ contains
     energy         =orbit%energy         ()
     angularMomentum=orbit%angularMomentum()
     ! Compute velocity components.
-    newVelocityTangential=angularMomentum/newRadius
+    newVelocityTangential=angularMomentum/newRadius/orbit%specificReducedMass()
     newVelocityRadial    =sqrt(2.0d0*(energy+gravitationalConstantGalacticus*orbit%massHost()/newRadius)/orbit%specificReducedMass()-newVelocityTangential**2)
     ! Move to the infalling phase of the orbit if requested.
     if (present(infalling)) then


abensonca:
This fix looks good. It breaks the [deterministic spins constraint](https://github.com/galacticusorg/galacticus/actions/runs/6018740185/job/16329338720?pr=454) check however. This is presumably because the fix to the orbit propagation slightly changes the angular momenta of merging halos, which are used to compute the spins. I will re-run that model and see how much the results are changed. Then we can either recalibrate the deterministic spins model, or just change the tolerance on the test to accept the new result.

abensonca:
The shift in the spin distribution is quite significant:
image
So, I'll attempt to recalibrate it (approximately).

Xiaolong-Du:
> The shift in the spin distribution is quite significant:
> So, I'll attempt to recalibrate it (approximately).

Agreed, the change in the spin distribution is significant. Since the tangential velocity is increased by a factor of $1+M_{\rm sat}/M_{\rm host}$ after the fix, the spin parameter $\lambda$ is probably also increased roughly by the same factor.

abensonca:
I've add a fix for this. The `nodeOperatorHaloAngularMomentumVitvitska2002` class already contains a parameter `massExponent` which scales the results with $1+M_\mathrm{sat}/M_\mathrm{host}$ so it should be as simple as adjusting this to offset this factor in the orbit propagation.

abensonca (approved):
The fix looks good. OK to commit once all tests pass.

PR #455, opened 2023-08-30
Use series solution for the radial velocity dispersion at large radii

Xiaolong-Du:
For certain combinations of $\alpha$, $\beta$ and $\gamma$, the radial velocity dispersion of the `darkMatterProfileDMOZhao1996` class can be computed analytically. However, the calculations can be inefficient. More importantly, at large radii, computing the analytic formula with machine accuracy may produce large round off errors and sometimes lead to floating-point errors. To solve this issue, we can use series solution at large radii.


abensonca (approved):
Everything looks good. I pushed one update with minor formatting changes. OK to commit once all tests pass.

Xiaolong-Du:
> I also ported this changes over to the [`massDistributions`](https://github.com/galacticusorg/galacticus/commit/2900b6d617332d71d2e304648eb032cc630195a6) branch so that they will be retained once that refactoring of how mass distributions are handled is completed.

Sounds good, thanks!

PR #478, opened 2023-09-21
Fix nBody operator

Xiaolong-Du:
Fix two issues:
1. When doing self-bound analysis, the most-bound particle in velocity space is found by searching for the minimum of potential in velocity space (similar to the calculation of gravitational potential in position space). In the current version of Galacticus, the softening velocity (similar to the definition of softening length) is fixed at 1 km/s by default. But this default value may be too large for small halos with typical particle velocity smaller than 1km/s. Instead, we can define a softening velocity as
$v_{\rm softening} = \sqrt{G m_p/r_{\rm softening}},$
where $m_p$ is the particle mass, $r_{\rm softening}$ is the softening length.

2.  When computing the gravitational potential at particle positions, a "omp do" directive is missing, which lead to incorrect results.

### DIFF
--- a/source/nBody.operator.potential_energy.F90
+++ b/source/nBody.operator.potential_energy.F90
@@ -193,6 +193,7 @@ contains
              ! Build octree.
              call octreePosition%build(positionRescaled,dble(selfBoundStatus(:,i)))
              !$omp parallel private(j)
+             !$omp do schedule(dynamic)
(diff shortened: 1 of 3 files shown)


abensonca (approved):
Everything looks good. Using this adaptive velocity softening is a good idea.

PR #483, opened 2023-09-28
Improvements to the Radial Range that Subhalo Density Profiles are Tabulated when Fitting TNFW Profile

cgannonucm:
Changed the range of radii at which the density profile is tabulated at when fitting TNFW profiles. Before the density was tabulated to the outer radius (where the radius encloses the bound mass or the total mass (of the subhalo), whichever is smaller). This definition led to the maximum radius taking on very large values in some cases. To fix this, the following changes have been implemented: 
- Now the maximum tabulated radius is the virial radius if the outer radius is less than the virial radius
- If the outer radius is greater than the virial radius, the maximum radius is the radius at which the density drops to 1/10 the density at the virial radius or 10x the virial radius, whichever is smaller

### DIFF
--- a/source/nodes.property_extractor.tidally_truncated_NFW_fit.F90
+++ b/source/nodes.property_extractor.tidally_truncated_NFW_fit.F90
@@ -165,18 +165,21 @@ contains
     type            (multiCounter                               ), intent(inout), optional     :: instance
     class           (nodeComponentDarkMatterProfile             ), pointer                     :: darkMatterProfile
     class           (nodeComponentSatellite                     ), pointer                     :: satellite
-    double precision                                             , allocatable  , dimension(:) :: radii                     , fractionDensity
+    double precision                                             , allocatable  , dimension(:) :: radii                                          , fractionDensity
     type            (multiDMinimizer                            ), allocatable                 :: minimizer_
     double precision                                                            , dimension(1) :: locationMinimum
-    double precision                                             , parameter                   :: fractionRadiusOuter=0.95d0, fractionRadiusScale=0.1d0, &
-         &                                                                                        fractionMaximum    =0.10d0, fractionStep       =0.1d0
-    integer                                                      , parameter                   :: countRadiiPerDecade=10
-    integer                                                                                    :: countRadii                , i                        , &
-         &                                                                                        iteration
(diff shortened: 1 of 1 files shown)


source/nodes.property_extractor.tidally_truncated_NFW_fit.F90
```diff
@@ -190,8 +192,18 @@ function tidallyTruncatedNFWFitExtract(self,node,time,instance)
        radiusScale  =  darkMatterProfile                     %scale              (                                              )
        radiusVirial =  self             %darkMatterHaloScale_%radiusVirial       (node                                          )
        ! Choose radii for fitting.
-       radiusMaximum=    fractionRadiusOuter*radiusOuter
-       radiusMinimum=min(fractionRadiusScale*radiusScale,fractionMaximum*radiusMaximum)
+       radiusMaximum=radiusVirial
+       if (radiusOuter > radiusVirial) then
+           radiusMaximumCountRadii=int(log10(radiusMaximumScaleVirialMaximum)*dble(radiusMaximumCountRadiiPerDecade)+1.0d0)
+           factorStepRadius       =    log10(radiusMaximumScaleVirialMaximum)/dble(radiusMaximumCountRadii         ) 
+           do i=1,radiusMaximumCountRadii
```
abensonca, line 199: This was previously:
```
         do n=1,radiusMaxiumCountRadii
             radiusMaximum                     =10**(log10(radiusVirial)+step*dble(i))
```
So the loop variable was `n` but on the following line it used `i` to increment `radiusMaximum`. Most likely `i=0` initially, so this would not have actually increased `radiusMaximum`, leaving it always equal to `radiusVirial`. I've corrected this (just using `i` as the loop variable), but it would be good if you can check if this still gives good results for the fits.

abensonca (commented):
Overall this looks good - I saw one issue with a difference between the name of a loop variable and the variable actually used in the loop (see other comment). I've fixed this, but let's check if the results still look good after this change.

source/nodes.property_extractor.tidally_truncated_NFW_fit.F90
cgannonucm: I recompiled and re-ran the parameter with the bug fixed. The fits look good.

abensonca (approved):
All tests have already passed, and since the fits still look good with the previous fix, this is ready to merge.

PR #493, opened 2023-10-09
Changes to accretion.halo.cold_mode.F90 and cooling.cooling_function.molecular_hydrogen_Galli_Palla.F90

sachiwee:
Added a temperatureCold to accretion.halo.cold_mode.F90
Set minimum temperature checks for cooling.cooling_function.molecular_hydrogen_Galli_Palla.F90


abensonca (approved):
These fixes all look good.

PR #572, opened 2024-03-02
fix: Use series expansion in computing radius enclosing a small mass for NFW profile.

Xiaolong-Du:
When computing the radius enclosing a tiny mass for NFW profile, the analytic formula is inaccurate due to roundoff errors, which returns effectively zero value. This may lead to floating-point errors when using NFW profile in the heated profile. To avoid this issue, we can make use of the series expansion of the r(M) function.


abensonca (approved):
This looks good - thanks for the improvement!

PR #583, opened 2024-03-15
Orbit calibration

Xiaolong-Du:
Implement an improved model for tidal evolution described in

http://arxiv.org/abs/2403.09597


abensonca (approved):
I made some minor formatting fixes. Otherwise all looks good.

PR #585, opened 2024-03-18
Add new HII region luminosity calculations

sachiwee:
Add new calculations to calculate HII region luminosities by summing over SFHs


abensonca (approved):
I cleaned up some formatting and renamed some files. Everything looks good now.

PR #613, opened 2024-05-16
Added nodePropertyExtractorSatelliteDynamicalTime

cgannonucm:
Added nodePropertyExtractorSatelliteDynamicalTime, a satellite dynamical time extractor class

### DIFF
--- a/source/nodes.property_extractor.satellite.dynamical_time.F90
+++ b/source/nodes.property_extractor.satellite.dynamical_time.F90
@@ -1,0 +1,189 @@
+!! Copyright 2009, 2010, 2011, 2012, 2013, 2014, 2015, 2016, 2017, 2018,
+
+!!           2019, 2020, 2021, 2022, 2023, 2024
+!!    Andrew Benson <abenson@carnegiescience.edu>
+!!
+!! This file is part of Galacticus.
+!!
+!!    Galacticus is free software: you can redistribute it and/or modify
+!!    it under the terms of the GNU General Public License as published by
+!!    the Free Software Foundation, either version 3 of the License, or
+!!    (at your option) any later version.
+!!
+!!    Galacticus is distributed in the hope that it will be useful,
+!!    but WITHOUT ANY WARRANTY; without even the implied warranty of
+!!    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+!!    GNU General Public License for more details.
+!!
+!!    You should have received a copy of the GNU General Public License
+!!    along with Galacticus.  If not, see <http://www.gnu.org/licenses/>.
+
+
+!!{
+Contains a module which implements a satellite dynamical time extractor class
+!!}
+
+  use :: Satellite_Tidal_Stripping_Radii,   only : satelliteTidalStrippingRadiusClass
+  use :: Galactic_Structure             ,   only : galacticStructureClass
+
+  !![
+  <nodePropertyExtractor name="nodePropertyExtractorSatelliteDynamicalTime">
+   <description>
+    A satellite dynamical time extractor class. Extracts the satellite dynamical time in Gyr
+   </description>
+  </nodePropertyExtractor>
+  !!]
+  type, extends(nodePropertyExtractorScalar) :: nodePropertyExtractorSatelliteDynamicalTime
+     !!{
+     A dynamical time extractor class.
+     !!}
+     private
+     class(satelliteTidalStrippingRadiusClass), pointer :: satelliteTidalStrippingRadius_ => null()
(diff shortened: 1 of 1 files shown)


abensonca (commented):
I made some minor clean ups to the code.

One larger question. You have:
$$\tau_\mathrm{dyn} = \sqrt{\frac{3 \pi}{16 \mathrm{G} \rho_\mathrm{tidal}}} = \sqrt{\frac{3 \pi r_\mathrm{tidal}^3}{16 \mathrm{G} M_\mathrm{tidal}}},$$
But, I assume that the density should be estimated as
$$\rho_\mathrm{tidal} = \frac{3 M_\mathrm{tidal}}{4 \pi r_\mathrm{tidal}^3},$$
so we would get
$$\tau_\mathrm{dyn} = \sqrt{\frac{3 \pi}{16 \mathrm{G} \rho_\mathrm{tidal}}} = \sqrt{\frac{\pi^2 r_\mathrm{tidal}^3}{4 \mathrm{G} M_\mathrm{tidal}}}.$$
I made this change in the code - can you check through and see if this is correct (both mathematically and in the code)?

abensonca:
> I made some minor clean ups to the code.
> 
> One larger question. You have: τdyn=3π16Gρtidal=3πrtidal316GMtidal, 

Also, I'm assuming you're taking this definition from Binney & Tremaine? I added a citation for this in the code. If you're using some other source we can cite that instead,

cgannonucm:
> > I made some minor clean ups to the code.
> > One larger question. You have: τdyn=3π16Gρtidal=3πrtidal316GMtidal,
> 
> Also, I'm assuming you're taking this definition from Binney & Tremaine? I added a citation for this in the code. If you're using some other source we can cite that instead,



> > I made some minor clean ups to the code.
> > One larger question. You have: τdyn=3π16Gρtidal=3πrtidal316GMtidal,
> 
> Also, I'm assuming you're taking this definition from Binney & Tremaine? I added a citation for this in the code. If you're using some other source we can cite that instead,



> > I made some minor clean ups to the code.
> > One larger question. You have: τdyn=3π16Gρtidal=3πrtidal316GMtidal,
> 
> Also, I'm assuming you're taking this definition from Binney & Tremaine? I added a citation for this in the code. If you're using some other source we can cite that instead,

Hi Andrew, I was using the definition given in Xiaolong et al (2024).

abensonca:
@Xiaolong-Du Can you comment on the origin of the dynamical time equation from [Du et al. (2024)](https://arxiv.org/abs/2403.09597) - specifically the factors of 3, 16, $\pi$, etc.? See the above discussion with Charles. The $\sqrt{3 \pi / 16 \mathrm{G} \rho$ definition looks like the one from Binney & Tremaine, but then I would expect $\sqrt{\pi^2 r^3 / 4 \mathrm{G} M$ when expressed in terms of radius and mass. At some level it doesn't matter - the numerical factors in this are just a matter of definition - but it would be good to be clear about where these originate from.

Xiaolong-Du:
> @Xiaolong-Du Can you comment on the origin of the dynamical time equation from [Du et al. (2024)](https://arxiv.org/abs/2403.09597) - specifically the factors of 3, 16, π, etc.? See the above discussion with Charles. The 3π/16Gρ definition looks like the one from Binney & Tremaine, but then I would expect π2r3/4GM when expressed in terms of radius and mass. At some level it doesn't matter - the numerical factors in this are just a matter of definition - but it would be good to be clear about where these originate from.

@abensonca You are right that the dynamical time equation in [Du et al. (2024)](https://arxiv.org/abs/2403.09597) is from Binney & Tremaine. When I compute the mean density, I have neglected an order-1 factor. Maybe we can change this slightly in the paper to match the exact definition in Binney & Tremaine.

abensonca:
@Xiaolong-Du OK, sounds good. In that case I suggest that we keep the modified version:
$$\tau_\mathrm{dyn} = \sqrt{\frac{3 \pi}{16 \mathrm{G} \rho_\mathrm{tidal}}} = \sqrt{\frac{\pi^2 r_\mathrm{tidal}^3}{4 \mathrm{G} M_\mathrm{tidal}}}.$$
that I have implemented in Charles new code.

If we do change the definition in your paper I guess we would need to make a corresponding change in the tidal mass loss rate class?

Xiaolong-Du:
Yes, we will need to modify the tidal mass loss rate class correspondingly. We will also need to change the best-fit parameter in the test suites, i.e. dividing the stripping coefficient by a factor of $\sqrt{4 \pi / 3}$.

abensonca:
Sounds good - thanks!

abensonca (approved):
Everything looks good now.

PR #614, opened 2024-05-18
fix: floating point exception in nodePropertyExtractorSatelliteDynamicalTime

cgannonucm:
Fixed floating point exception in nodePropertyExtractorSatelliteDynamicalTime. Check that both the tidal radius and tidal mass are non zero. If they are zero, return -1.0 to avoid divide by zero.


abensonca (approved):
All looks good - thanks for the fix.

PR #615, opened 2024-05-29
fix: Remove `Method` suffix from parameter names in test suite parameter files

abensonca:
This should have happened along with the [commit](https://github.com/galacticusorg/galacticus/commit/f483d9caf532e5f33cdbe46aad53d076316ec394) that first removed these suffixes, but these two files were apparently missed.


Xiaolong-Du:
@abensonca For this dark matter-only case, it might also be helpful to set `OmegaBaryon=0`. In some tests, I find that with a nonzero `OmegaBaryon`, I got floating-point errors, which seems to be related to resetting the obit of satellite when a subhalo is promoted.

abensonca:
Good idea. I've changed the parameter files so that baryons are included in the calculation of the transfer function (so that this matches what was done in the COCO simulations to which this model was calibrated), but are _not_ included in the rest of the calculation.

PR #627, opened 2024-06-12
Add summation over HII region n_H densities

sachiwee:
Update the calculation of emission line luminosities in HII regions to sum over a distribution of hydrogen densities.


abensonca (approved):
Everything looks good now - will commit after tests pass.

PR #628, opened 2024-06-14
Output analyses class for black hole mass- velocity dispersion relation

sachiwee:
Output analyses class for black hole mass- velocity dispersion relation using the observations of McConnel & Ma 2013.


abensonca (approved):
I made some formatting fixes and added the citation to the bibliography. All looks good now - will merge after tests pass.

PR #647, opened 2024-07-16
Change the definition of dynamical timescale to follow that in Binney & Tremaine, "Galactic Dynamics".

Xiaolong-Du:
A constant of order unity was ignored in the definition of dynamical timescale, $T_{\rm dyn}$, when computing the tidal mass loss. To be consistent with the definition in the book "Galactic Dynamics" by Binney & Tremaine, $T_{\rm dyn}$ is changed from
$$T_{\rm dyn} = \sqrt{\frac{3 \pi r_{\rm t}^3}{16  G  M_{\rm sub}(r_{\rm t})}}$$
to
$$T_{\rm dyn} = \sqrt{\frac{3 \pi}{16 G  \overline{\rho}(r_{\rm t})}} = 2 \pi \sqrt{\frac{r_{\rm t}^3}{16  G  M_{\rm sub}(r_{\rm t})}}.$$
The updated definition gives a value $\sqrt{4 \pi /3}$ times larger than that from the old definition. Accordingly, the best-fit value of the tidal stripping coefficient $\alpha_{s}$ must also be $\sqrt{4 \pi /3}$ times larger in order to keep the tidal mass loss unchanged, i.e.
$$\frac{d M_{\rm sat}}{dt} = -\alpha_s \frac{M_{\rm sat} - M(r_{\rm t})}{T_{\rm dyn}}.$$

This issue is related to #613. @cgannonucm

### DIFF
--- a/source/satellites.tidal_stripping.rate.Zentner2005.F90
+++ b/source/satellites.tidal_stripping.rate.Zentner2005.F90
@@ -50,7 +50,7 @@
     If {\normalfont \ttfamily [useDynamicalTimeScale]} is set to true, $T_\mathrm{loss}$ is taken to be the dynamical time scale
     computed at the tidal radius
     \begin{equation}
-    T_\mathrm{dyn} = \sqrt{\frac{3\pi r_\mathrm{tidal}^3}{16 G M_\mathrm{sat}(r_\mathrm{tidal})}}.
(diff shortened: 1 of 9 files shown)


abensonca (approved):
Everything looks good - thanks for the fix!

abensonca (approved):

PR #669, opened 2024-08-19
Fix lightcone redshift observed

Andrew-Robertson:
Peculiar and cosmological redshifts should add according to $1 + z_\mathrm{obs} = (1 + z_\mathrm{cos}) (1 + z_\mathrm{pec})$. The old code contained a bug where $z_\mathrm{obs} = z_\mathrm{cos} + (1 + z_\mathrm{cos}) (1 + z_\mathrm{pec})$. This has been fixed. 

In addition, the code previously used a special relativistic formula for $z_\mathrm{pec}$, valid when the peculiar velocity is purely along the line-of-sight. A full special-relativistic treatment would include time dilation effects (leading to a non-zero $z_\mathrm{pec}$ even for motion in the plane of the sky), but also shifts in the positions (on the sky) where galaxies are observed, due to relativistic aberration. Given these other relativistic effects are being ignored, the code now uses the non-relativistic limit in which $z_\mathrm{pec} = v_\mathrm{los} / c$, with $v_\mathrm{los}$ the line-of-sight component of the peculiar velocity.

### DIFF
--- a/source/nodes.property_extractor.lightcone.F90
+++ b/source/nodes.property_extractor.lightcone.F90
@@ -274,7 +274,7 @@ contains
          &                                                                          )
     lightconeExtract   (8  )=self%geometryLightcone_%solidAngle()/degreesToRadians**2
     if (self%includeObservedRedshift) then
-       ! Compute the relativistic velocity β=v/c.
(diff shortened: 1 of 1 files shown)


abensonca (approved):
All looks good - thanks for the fix!

PR #670, opened 2024-08-20
Add stream_kick.F90 filter for stream velocity kick.

pmenker:
Works in circular case.


abensonca (approved):
I did some general clean up - all looks good now.

PR #709, opened 2024-09-25
fix(perf): only load the necessary parts of each dataset

Andrew-Robertson:
Reduces the required memory and speeds up the extraction of a single merger tree from a merger tree file.


abensonca (approved):
All looks good - thanks for the fix!

PR #732, opened 2024-10-29
Implement a merger tree build controller class which builds trees containing only the main branch and certain descendants

Xiaolong-Du:
Sometimes, we do not need to build the full merger trees, e.g. we want to look at the properties of the halos on the main branch. In such cases, building a branchless merger tree will help to significantly speed up the tree building process and save the memory needed to store the tree.

However, some properties of the halos on the main branch may depend on the detailed merger history. For example, if one uses the `ludlow2016` model to determine the halo concentrations, the concentration of a halo will depends on all its progenitors above a certain mass fraction, i.e. not only the progenitors on the main branch. In this case, a branchless merger tree can not be used. To solve this problem, we can keep the nodes on the main branch and all its descendants above a specific mass fraction. This will still significantly speed up calculations and reduce the memory cost, but make sure that the concentration of halos on the main branch can be correctly determined from the tree.

### DIFF
--- a/source/merger_trees.build.controller.main_branch.F90
+++ b/source/merger_trees.build.controller.main_branch.F90
@@ -1,0 +1,187 @@
+!! Copyright 2009, 2010, 2011, 2012, 2013, 2014, 2015, 2016, 2017, 2018,
+!!           2019, 2020, 2021, 2022, 2023, 2024
+!!    Andrew Benson <abenson@carnegiescience.edu>
+!!
+!! This file is part of Galacticus.
+!!
+!!    Galacticus is free software: you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


Xiaolong-Du:
Here are some comparisons between the full merger tree building algorithm and the new merger tree build controller. The first plot shows the concentration distribution of halos on the main branch at different redshifts determined using the `ludlow2016` model. The root node is at $z=0.5$  and has a mass of $\sim 10^{13} M_{\odot}$. The second and third plots show the time and peak memory costs, respectively.
c_200_compare
Timing
Memory

abensonca (approved):
I made a few minor style changes - everything else looks good. Will commit after all tests pass.

PR #766, opened 2025-01-13
Update copyright year

abensonca:
Update copyright year to 2025.


abensonca:
Ignoring the failure of the LaTeX-checker - it just times out given the huge number of files it attempts to process.

PR #767, opened 2025-01-16
Update Shen 2003 mass-size relation to reflect erratum

Andrew-Robertson:
Erratum is here: https://ui.adsabs.harvard.edu/abs/2007MNRAS.379..400S/abstract


abensonca (approved):
Looks good - thanks for the fix!

PR #782, opened 2025-02-04
The central density of the β-profile was evaluated at the wrong radius in cooling radius calculations

abensonca:
This was being evaluated at the outer radius instead of at zero radius.


liempi:
> If this looks correct to you, you can go ahead and approve in the review, and I'll then commit the fix.

@abensonca green light from my side! Thanks for solving this issue.

PR #805, opened 2025-02-27
Support parameter change files

abensonca:
Allows specifying one or more files which define changes to be applied to the primary parameter file. Documentation [here](https://github.com/galacticusorg/galacticus/wiki/Tutorial:-Introduction-to-Galacticus-parameter-files#changing-parameters).


abensonca:
Also updates the way `modelParameter` names are specified in MCMC config files to follow [XPath](http://www.w3schools.com/Xml/xpath_syntax.asp) conventions, and to allow addressing by attribute values. For example, a valid name is now:
```
nodeOperator/nodeOperator[@value='stellarFeedbackDisks']/stellarFeedbackOutflows/stellarFeedbackOutflows/velocityCharacteristic
```
where:

1. The old `::` separator is replaced with the `/` XPath separator;
2. We can now use `nodeOperator[@value='stellarFeedbackDisks']` to select the `nodeOperator` by choosing the one with a given `value` attribute.

Note that the old numeric indexing still works, so:
```
nodeOperator/nodeOperator[3]/stellarFeedbackOutflows/stellarFeedbackOutflows/velocityCharacteristic
```
is still valid, **_but_** the indexing now starts from 1 instead of from 0 (to conform to XPath standards). This is necessary as currently we only support a very limited set of XPath functionality which means that there are some circumstances where addressing via index remains the only viable option.

The `parametersMigrate.pl` script is aware of these changes, and will automatically change `::` to `/` and increment any index values by 1.

PR #809, opened 2025-03-12
Improve nodes.property_extractor class for HII regions in galaxies

sachiwee:
-  New class to implement a fixed escape fraction given age of stellar population.
-  Rearrange and slicing  luminosity array from Cloudy table as expected by the rest of the code (needs Cloudy table with       indices stored in attributes of "metallicity" dataset)
- Remove "if (     present(node       ).or.present(starFormationHistory)) call Error_Report('`node` is not supported'  //{introspection:location})" from  star_formation.histories.adaptive.F90


abensonca (approved):
I made a few changes to improve formatting and make some variable names more descriptive. Otherwise, everything looks good and I'll merge this once tests have passed.

PR #818, opened 2025-03-21
Extractor for emission line luminosities from narrow line regions of galaxies

sachiwee:
This nodes.property_extractor class computes narrow line region luminosities of galaxies by interpolating galaxy’s metallicity, and the AGN's ionization parameter, and spectral index from the Cloudy grid assuming constant density 100 cm^-3


abensonca (approved):
This all looks good. I cleaned up some formatting and variable names. 

Two things for you to check.

1. The variable `halfMassRadius` was computed but not actually used anywhere. So I removed it, but I wanted to check if this is correct - was it supposed to be used somewhere in the calculation?

2. I used the `atomicRecombinationRateRadiative` class to get the hydrogen recombination coefficient. The value should be very close to what you used, but check that the results haven't changed too much.

If that all seems ok I'll merge this once tests have passed.

sachiwee:
Hi Andrew,
Thank you so much!
1. Yes, I think you suggested we could use half Mass radius as an approximate to stronmgren radius when we were still in the early stages. We don’t need it anymore. Thanks so much for removing it.
2. Great! I will check this and let you know!
Thanks so much!

PR #822, opened 2025-04-05
fix: Correct nuclear star cluster angular momentum rate and black hole seed mass 

liempi:
In the `source/nodes.operators.physics.nuclear_star_clusters.growth.F90` node operator, we were removing angular momentum from the nuclear star cluster instead of adding it from the spheroid.

Additionally, the file `source/black_holes.seeds.Vergara2023.F90` contained unused modules and variables. We also adjusted the stellar mass of the nuclear star cluster before calculating the black hole seed mass, resulting in an incorrect calculation of the initial black hole seed mass.


abensonca (approved):
All looks good. I'll merge this once tests pass.

PR #852, opened 2025-05-23
Reduce the Chandrasekhar integral (for dynamical friction) in spherical systems to account for the finite size of the perturber

abensonca:
Uses a simple (largely untested) model in which the integral is reduced by a factor $f = M(r_\mathrm{orb})/M$ where $M(r)$ is the mass profile of the perturber, $M$ is its total mass, and $r_\mathrm{orb}$ is its orbital radius in its host.

### DIFF
--- a/source/mass_distributions.spherical.F90
+++ b/source/mass_distributions.spherical.F90
@@ -945,7 +945,7 @@ contains
     return
   end function sphericalSurfaceDensity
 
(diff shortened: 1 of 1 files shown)


abensonca:
Need to check carefully for any change in validation metrics (for dark matter only subhalo models in particular) before merging this PR.

PR #853, opened 2025-05-27
Implement FDM profile based on Chowdhury2021

yzhaoastro:
This PR adds a new massDistribution and corresponding darkMatterProfile implementation for modeling FDM halos. The test file is used to generate the density profile of FDM halos following Chowdhury et al. (2021). If the parameters of the FDM halo are not provided explicitly, the code applies the core-halo relation from Schive et al. (2014) to compute them.


abensonca (approved):
Everything looks good and all tests passed. Will merge.

PR #871, opened 2025-06-28
Add analytic solution for soliton and NFW mass profile

yzhaoastro:
At small radii, a third-order Taylor expansion (retaining terms up to r^7) is applied to improve numerical stability. The relative error of this approximation is estimated to be below 10^{-6} at r/r_c=0.1, using FDM halo parameters adopted in Chowdhury et al. (2021).

### DIFF
--- a/source/mass_distributions.spherical.solition_NFW.F90
+++ b/source/mass_distributions.spherical.solition_NFW.F90
@@ -42,6 +42,7 @@
           &              radiusVirial           , radiusCoreScaleFree    , &
           &              radiusSolitonScaleFree , densitySolitonScaleFree
    contains
+     procedure :: mass                  => solitonNFWMass
      procedure :: density               => solitonNFWDensity
      procedure :: densityGradientRadial => solitonNFWDensityGradientRadial
      procedure :: parameters            => solitonNFWParameters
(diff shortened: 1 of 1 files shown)


source/mass_distributions.spherical.solition_NFW.F90
```diff
@@ -250,6 +251,65 @@ function solitonNFWFactoryTabulation(self,parameters) result(instance)
       return
    end function solitonNFWFactoryTabulation
 
+   double precision function SolitonMass(self,r) result(f)
```
abensonca, line 254: Function names should typically start with lower case letters. The result should have a more descriptive name, e.g. `mass` instead of `f`.

source/mass_distributions.spherical.solition_NFW.F90
```diff
@@ -250,6 +251,65 @@ function solitonNFWFactoryTabulation(self,parameters) result(instance)
...
+    !!{
+    Return the mass at the specified {\normalfont \ttfamily coordinates} in the soliton region.
+    !!}
+    use :: Numerical_Constants_Math        , only : Pi
+    implicit none
+    class           (massDistributionSolitonNFW), intent(inout) :: self
+    double precision                            , intent(in   ) :: r
+    double precision                            , parameter     :: A = 0.091d0
```
abensonca, line 262: It's good to define this factor of 0.091 as a consent. Since it's used in the density function also, you could move this definition into the submodule scope (around line 64) so that it's available to use here and in the density function also.

source/mass_distributions.spherical.solition_NFW.F90
```diff
@@ -250,6 +251,65 @@ function solitonNFWFactoryTabulation(self,parameters) result(instance)
...
+    double precision                                            :: radiusCore, densitySolitonCentral
+    double precision                                            :: atanterm  , prefactor            , term, poly
+
+    radiusCore           =+self%radiusCore
```
abensonca, line 266: There's no need to copy these values out of `self` here - it's ok to just use them directly.

source/mass_distributions.spherical.solition_NFW.F90
```diff
@@ -250,6 +251,65 @@ function solitonNFWFactoryTabulation(self,parameters) result(instance)
...
+      Return the mass at the specified {\normalfont \ttfamily coordinates} in a soliton and NFW mass distribution.
+      !!}
+      use :: Numerical_Constants_Math        , only : Pi
+      implicit none
+      class           (massDistributionSolitonNFW), intent(inout) :: self
+      class           (coordinate                ), intent(in   ) :: coordinates
+      double precision                                            :: Msol       , invterm, logterm, r
+
+      r         =+coordinates%rSpherical()
+      Msol      =+SolitonMass           (self, self%radiusSoliton)
```
abensonca, line 300: I think you can move this into the "NFW regime" section as it's only needed there. Then we avoid the cost of computing it in the soliton regime.

abensonca (changes requested):
Thanks for the update! I left a comments in the code. Other than those it would be good to use more descriptive variable names in a few places. For example, `radius` instead of `r`, `massSoliton` instead of `MSol`. This makes it easier for other people to understand the code (for example sometime might think that `MSol` means Solar masses).

 I saw that the documentation build failed. I'm looking into why that happened.

abensonca:
I fixed the problem with the documentation build. The new method for computing the mass needs to match the name of the overridden method in the parent class, so I changed it to `massEnclosedBySphere`.

yzhaoastro:
Thanks! I've renamed the function to `massEnclosedBySphere` to match the parent class and updated the variable names for clarity.

abensonca (approved):
Everything looks good now. Will merge this once all tests pass.

PR #882, opened 2025-07-16
Add a kinematics class for the Soliton-NFW profile

yzhaoastro:
This PR adds a new kinematics distribution class for the soliton+NFW mass profile:

Implements kinematicsDistributionSolitonNFW in `kinematic_distributions.collisionless.soliton_NFW.F90`.

Uses analytic scaling at small radii to avoid tabulating velocity dispersion where the profile becomes power-law. Specifically, since the soliton core behaves as ρ ~ r^0, the velocity dispersion scales as σ ~ r^1 at small radii.

This improves tabulation speed and numerical stability, especially when tidal heating calculations request values at very small radii.

In addition, this PR fixes a bug in the `dark_matter_profiles_DMO.soliton_NFW.F90` implementation where the rootfinder's `rangeExpand` was not being re-run after `radiusCore` was updated.

### DIFF
--- a/source/dark_matter_profiles_DMO.soliton_NFW.F90
+++ b/source/dark_matter_profiles_DMO.soliton_NFW.F90
@@ -47,7 +47,7 @@
      class           (cosmologyFunctionsClass   ), pointer :: cosmologyFunctions_                => null()
      class           (darkMatterParticleClass   ), pointer :: darkMatterParticle_                => null()
      class           (virialDensityContrastClass), pointer :: virialDensityContrast_             => null()
-     double precision                                      :: toleranceRelativeVelocityDispersion        , toleranceRelativeVelocityDispersionMaximum
+     double precision                                      :: toleranceRelativeVelocityDispersion         , toleranceRelativeVelocityDispersionMaximum
    contains
      !![
      <methods>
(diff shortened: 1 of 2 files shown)


abensonca (approved):
I made one fix (and a few other minor formatting adjustments). You originally had:
```
             radiusScaleFreeSmall=+fractionSmall                              &        
                  &               *(                                          &                                                                                                                                                  
                  &               +massDistributionEmbedding%radiusCore       & 
                  &               /massDistributionEmbedding%radiusScale      &                 
                  &                )**2                                                                         
```
 I think the use of the square of the core radius here isn't necessary (this is probably left over from the implementation for the cusp-NFW profile where we have to compare the radius to $y^2$ since, in that profile, the relevant term in the profile is $y^2 + (r/r_\mathrm{s})^2$. So, I replaced this with:
```
             radiusScaleFreeSmall=+fractionSmall                           &                             
                  &               *massDistributionEmbedding%radiusCore    &        
```

I'll merge this when all tests pass.

yzhaoastro:
Thanks for the fix!

abensonca:
The build is failing with an error:
```
Error: Component 'darkmatterparticle_' at (1) is a PRIVATE component of 'darkmatterprofiledmosolitonnfw'
```
This is probably because the "internal" constructor for the `darkmatterprofiledmosolitonnfw` class was changed (to add the new tolerance parameters). So we'll need to update the call to this constructor in the test code for the fuzzy dark matter profile also. You can just set some default values for the tolerances in that test code and then check that it still runs successfully.

PR #898, opened 2025-08-13
Add a radiusEnclosingDensity function and scatter to core-Mass relation for the Soliton-NFW profile 

yzhaoastro:
Added radiusEnclosingDensity function to the Soliton-NFW profile.
Update Soliton-NFW profile with new core-mass relation and fix tests.dark_matter_profiles.fuzzy_dark_matter accordingly.
Updated core-mass relation based on Chan et al. 2022, [https://ui.adsabs.harvard.edu/abs/2022MNRAS.511..943C].

### DIFF
--- a/source/dark_matter_profiles_DMO.soliton_NFW.F90
+++ b/source/dark_matter_profiles_DMO.soliton_NFW.F90
@@ -23,11 +23,12 @@
   An implementation of fuzzy dark matter halo profiles using the soliton and NFW mass distribution.
   !!}
 
-  use :: Dark_Matter_Halo_Scales, only : darkMatterHaloScaleClass
-  use :: Dark_Matter_Particles  , only : darkMatterParticleClass
-  use :: Cosmology_Parameters   , only : cosmologyParametersClass
-  use :: Cosmology_Functions    , only : cosmologyFunctionsClass
-  use :: Virial_Density_Contrast, only : virialDensityContrastClass
+  use :: Dark_Matter_Halo_Scales , only : darkMatterHaloScaleClass
+  use :: Dark_Matter_Particles   , only : darkMatterParticleClass
+  use :: Cosmology_Parameters    , only : cosmologyParametersClass
+  use :: Cosmology_Functions     , only : cosmologyFunctionsClass
+  use :: Virial_Density_Contrast , only : virialDensityContrastClass
+  use :: Statistics_Distributions, only : distributionFunction1DNormal
   !![
   <darkMatterProfileDMO name="darkMatterProfileDMOSolitonNFW">
    <description>
@@ -41,14 +42,21 @@
      A dark matter halo profile class implementing \gls{fdm} dark matter halos.
      !!}
      private
-     double precision                                      :: massParticle
-     class           (darkMatterHaloScaleClass  ), pointer :: darkMatterHaloScale_               => null()
-     class           (cosmologyParametersClass  ), pointer :: cosmologyParameters_               => null()
-     class           (cosmologyFunctionsClass   ), pointer :: cosmologyFunctions_                => null()
-     class           (darkMatterParticleClass   ), pointer :: darkMatterParticle_                => null()
-     class           (virialDensityContrastClass), pointer :: virialDensityContrast_             => null()
-     double precision                                      :: toleranceRelativeVelocityDispersion         , toleranceRelativeVelocityDispersionMaximum
-     integer         (kind_int8                 )          :: lastUniqueID
+     double precision                                        :: massParticle
+     class           (darkMatterHaloScaleClass    ), pointer :: darkMatterHaloScale_               => null()
+     class           (cosmologyParametersClass    ), pointer :: cosmologyParameters_               => null()
+     class           (cosmologyFunctionsClass     ), pointer :: cosmologyFunctions_                => null()
+     class           (darkMatterParticleClass     ), pointer :: darkMatterParticle_                => null()
+     class           (virialDensityContrastClass  ), pointer :: virialDensityContrast_             => null()
+     type            (distributionFunction1DNormal)          :: alphaNormal
+     type            (distributionFunction1DNormal)          :: massCoreScatter
+     double precision                                        :: toleranceRelativeVelocityDispersion         , toleranceRelativeVelocityDispersionMaximum
+     double precision                                        :: radiusVirialPrevious                        , radiusScalePrevious                       , &
+          &                                                     radiusCorePrevious                          , radiusSolitonPrevious                     , &
+          &                                                     densityCorePrevious                         , densityScalePrevious                      , &
+          &                                                      massCorePrevious
+     integer          (kind_int8                  )          :: lastUniqueID
+     integer                                                 :: randomOffsetID                              , densityCoreID
    contains
      !![
      <methods>
@@ -141,9 +149,14 @@ contains
     double precision                                , intent(in)         :: toleranceRelativeVelocityDispersion, toleranceRelativeVelocityDispersionMaximum
     !![
     <constructorAssign variables="*darkMatterHaloScale_, *darkMatterParticle_, *cosmologyFunctions_, *cosmologyParameters_, *virialDensityContrast_, toleranceRelativeVelocityDispersion, toleranceRelativeVelocityDispersionMaximum"/>
+    <addMetaProperty component="basic" name="randomOffset" id="self%randomOffsetID" isEvolvable="no" isCreator="yes"/>
+    <addMetaProperty component="basic" name="densityCore"  id="self%densityCoreID"  isEvolvable="no" isCreator="yes"/>
     !!]
 
     self%lastUniqueID=-huge(1_kind_int8)
+    self%alphaNormal = distributionFunction1DNormal(mean=0.326d0,variance=1.0d-2)
+    self%massCoreScatter = distributionFunction1DNormal(mean=0.0d0,variance = log10(1.5d0)**2) ! 50% log-normal scatter from Eq.(15) of Chan et al. (2022; MNRAS; 551; 943; https://ui.adsabs.harvard.edu/abs/2022MNRAS.511..943C).
+
 
     select type (darkMatterParticle__ => self%darkMatterParticle_)
     class is (darkMatterParticleFuzzyDarkMatter)
@@ -180,6 +193,7 @@ contains
     !!{
     Destructor for the {\normalfont \ttfamily solitonNFW} dark matter halo profile class.
     !!}
+    use :: Events_Hooks, only : calculationResetEvent
     implicit none
     type(darkMatterProfileDMOSolitonNFW), intent(inout) :: self
 
@@ -190,6 +204,7 @@ contains
     <objectDestructor name="self%cosmologyParameters_"  />
     <objectDestructor name="self%virialDensityContrast_"/>
     !!]
+    if (calculationResetEvent%isAttached(self,solitonNFWCalculationReset)) call calculationResetEvent%detach(self,solitonNFWCalculationReset)
     return
   end subroutine solitonNFWDestructor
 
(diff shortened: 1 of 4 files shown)


source/dark_matter_profiles_DMO.soliton_NFW.F90
```diff
@@ -41,21 +42,32 @@
...
-     class           (darkMatterParticleClass   ), pointer :: darkMatterParticle_                => null()
-     class           (virialDensityContrastClass), pointer :: virialDensityContrast_             => null()
-     double precision                                      :: toleranceRelativeVelocityDispersion         , toleranceRelativeVelocityDispersionMaximum
+     double precision                                        :: massParticle
+     class           (darkMatterHaloScaleClass    ), pointer :: darkMatterHaloScale_               => null()
+     class           (cosmologyParametersClass    ), pointer :: cosmologyParameters_               => null()
+     class           (cosmologyFunctionsClass     ), pointer :: cosmologyFunctions_                => null()
+     class           (darkMatterParticleClass     ), pointer :: darkMatterParticle_                => null()
+     class           (virialDensityContrastClass  ), pointer :: virialDensityContrast_             => null()
+     type            (distributionFunction1DNormal)          :: alphaNormal
```
abensonca, line 51: `alphaNormal` seems not to be used - remove it?

source/dark_matter_profiles_DMO.soliton_NFW.F90
```diff
@@ -137,8 +149,14 @@ function solitonNFWConstructorInternal(darkMatterHaloScale_,darkMatterParticle_,
     double precision                                , intent(in)         :: toleranceRelativeVelocityDispersion, toleranceRelativeVelocityDispersionMaximum
     !![
     <constructorAssign variables="*darkMatterHaloScale_, *darkMatterParticle_, *cosmologyFunctions_, *cosmologyParameters_, *virialDensityContrast_, toleranceRelativeVelocityDispersion, toleranceRelativeVelocityDispersionMaximum"/>
+    <addMetaProperty component="basic" name="randomOffset" id="self%randomOffsetID" isEvolvable="no" isCreator="yes"/>
```
abensonca, line 152: Consider changing to `component="darkMatterProfile"` here. This isn't super important, but it's generally good to have meta-properties assigned to whichever component they are most closely related to.

source/dark_matter_profiles_DMO.soliton_NFW.F90
```diff
@@ -137,8 +149,14 @@ function solitonNFWConstructorInternal(darkMatterHaloScale_,darkMatterParticle_,
...
+    <addMetaProperty component="basic" name="densityCore"  id="self%densityCoreID"  isEvolvable="no" isCreator="yes"/>
     !!]
 
+    self%lastUniqueID=-huge(1_kind_int8)
+    self%alphaNormal = distributionFunction1DNormal(mean=0.326d0,variance=1.0d-2)
```
abensonca, line 157: `alphaNormal` is never used - remove?

source/dark_matter_profiles_DMO.soliton_NFW.F90
```diff
@@ -137,8 +149,14 @@ function solitonNFWConstructorInternal(darkMatterHaloScale_,darkMatterParticle_,
...
+    self%massCoreScatter = distributionFunction1DNormal(mean=0.0d0,variance = log10(1.5d0)**2) ! 50% log-normal scatter from Eq.(15) of Chan et al. (2022; MNRAS; 551; 943; https://ui.adsabs.harvard.edu/abs/2022MNRAS.511..943C).
```
abensonca, line 158: Maybe add to the `<description>` at line 35 to explain how this core radius is compute, and include a citation to Chan et al. there. If you add this:
```
@article{chan_diversity_2022,
	title = {The diversity of core-halo structure in the fuzzy dark matter model},
	volume = {511},
	issn = {0035-8711},
	url = {https://ui.adsabs.harvard.edu/abs/2022MNRAS.511..943C},
	doi = {10.1093/mnras/stac063},
	abstract = {In the fuzzy dark matter (FDM) model, gravitationally collapsed objects always consist of a solitonic core located within a virialized halo. Although various numerical simulations have confirmed that the collapsed structure can be described by a cored Navarro-Frenk-White-like density profile, there is still disagreement about the relation between the core mass and the halo mass. To fully understand this relation, we have assembled a large sample of cored haloes based on both idealized soliton mergers and cosmological simulations with various box sizes. We find that there exists a sizeable dispersion in the core-halo mass relation that increases with halo mass, indicating that the FDM model allows cores and haloes to coexist in diverse configurations. We provide a new empirical equation for a core-halo mass relation with uncertainties that can encompass all previously found relations in the dispersion, and emphasize that any observational constraints on the particle mass m using a tight one-to-one core-halo mass relation should suffer from an additional uncertainty of the order of 50 per cent for halo masses \$\{rsim\} 10{\textasciicircum}9 {\textbackslash}, [8{\textbackslash}times 10{\textasciicircum}\{-23\} {\textbackslash}, {\textbackslash}mathrm\{eV\}/(mc{\textasciicircum}2)]{\textasciicircum}\{3/2\} {\textbackslash}, {\textbackslash}mathrm\{M\}\_{\textbackslash}odot\$. We suggest that tidal stripping may be one of the effects contributing to the scatter in the relation.},
	urldate = {2025-08-13},
	journal = {Monthly Notices of the Royal Astronomical Society},
	author = {Chan, Hei Yin Jowett and Ferreira, Elisa G. M. and May, Simon and Hayashi, Kohei and Chiba, Masashi},
	month = mar,
	year = {2022},
	note = {Publisher: OUP
ADS Bibcode: 2022MNRAS.511..943C},
	keywords = {Astrophysics - Astrophysics of Galaxies, Astrophysics - Cosmology and Nongalactic Astrophysics, cosmology: theory, dark matter, galaxies: haloes, methods: numerical, software: simulations},
	pages = {943--952},
	file = {Full Text PDF:/home/abensonca/.mozilla/firefox/f54gqgdx.default/zotero/storage/6Q2EWPG5/Chan et al. - 2022 - The diversity of core-halo structure in the fuzzy .pdf:application/pdf},
}
```
to `docs/Galacticus.bib` then you can cite the paper with `\cite{chan_diversity_2022}` and it will be automatically included into the documentation.

source/mass_distributions.spherical.solition_NFW.F90
```diff
@@ -302,6 +303,10 @@ double precision function massSoliton(radius_) result(mass)
         double precision                :: termArcTangent           , termPrefactor , &
              &                             termCore                 , termPolynomial
         
+        if (radius_ <= 0.0d0) then
```
abensonca, line 306: This seems to warn that it will return zero, but then it doesn't actually do so! For `radius_=0` it should work anyway because the Taylor series used below will give zero.

Maybe it would be better to remove this warning, and in the `solitonNFWMassEnclosedBySphere()` function just add an extra `if`, e.g.:
```
      if (radius <= 0.0d0) then
         mass           =+0.0d0
      else if (radius < self%radiusSoliton) then
         ! Soliton regime.
         mass           =+massSoliton(radius)
      else
      .
      .
      .
```

source/mass_distributions.spherical.solition_NFW.F90
```diff
@@ -350,6 +355,10 @@ double precision function solitonNFWDensity(self,coordinates) result(density)
       class           (coordinate                ), intent(in   ) :: coordinates
       double precision                                            :: radiusScaleFree, radiusCoreFree
 
+      if (coordinates%rSpherical() <= 0.0d0) then
```
abensonca, line 358: This warning also seems to warn about something that doesn't happen. In this case, I would suggest removing it and modifying the `if` below to:
```
      if         (coordinates%rSpherical() <= 0.0d0) then
         ! Zero radius.
         density=+self%densitySolitonCentral
      else if (coordinates%rSpherical() < self%radiusSoliton) then
         ! Soliton regime.
         density=+self%densitySolitonCentral                  /(+1.0d0+coefficientCore*radiusCoreFree**2)**8
      else
         ! NFW regime.
         density=+self%densityNormalizationNFW/radiusScaleFree/(+1.0d0+                radiusScaleFree  )**2
      end if
```

source/mass_distributions.spherical.solition_NFW.F90
```diff
@@ -375,6 +384,10 @@ double precision function solitonNFWDensityGradientRadial(self,coordinates,logar
       <optionalArgument name="logarithmic" defaultsTo=".false."/>
       !!]
 
+      if (coordinates%rSpherical() <= 0.0d0) then
```
abensonca, line 387: Similar here - in this case just extend the `if` so that for r <= 0 it just returns a zero gradient.

source/mass_distributions.spherical.solition_NFW.F90
```diff
@@ -407,6 +420,24 @@ double precision function solitonNFWDensityGradientRadial(self,coordinates,logar
...
+    !!}
+    implicit none
+    class           (massDistributionSolitonNFW), intent(inout), target   :: self
+    double precision                            , intent(in   )           :: density
+    double precision                            , intent(in   ), optional :: radiusGuess
+
+    if (density >= self%densitySolitonCentral) then
+       radius=0.0d0
+    else
+       radius=sphericalTabulatedRadiusEnclosingDensity(self,density,radiusGuess)
```
abensonca, line 435: This is good, but can you check if:
```
      radius=self%massDistributionSphericalTabulated%radiusEnclosingDensity(self,density,radiusGuess)
```
works? I think it probably will not, in which case your approach here is ok.

abensonca (changes requested):
This all looks good. I left a few comments where we can clean up the code a little. I'll also push an update with a few minor formatting fixes.

source/dark_matter_profiles_DMO.soliton_NFW.F90
yzhaoastro: I’ve addressed all comments.

source/dark_matter_profiles_DMO.soliton_NFW.F90
```diff
@@ -305,37 +382,46 @@ subroutine solitonNFWComputeProperties(self,node,radiusVirial,radiusScale,radius
...
-         &                *hubbleConstant      &
-         &                *self%massParticle   &
-         &                /plancksConstantBar  &
-         &               )**1.5d0
-    massCore           =+0.25d0                                     & ! Equation (6) of Schive et al. (2014; PRL; 113; 1302; https://ui.adsabs.harvard.edu/abs/2014PhRvL.113z1302S).
-         &              *massHaloMinimum0                           &
-         &              /sqrt(expansionFactor)                      &
-         &              *(zeta_z  /zeta_0          )**(1.0d0/6.0d0) &
-         &              *(massHalo/massHaloMinimum0)**(1.0d0/3.0d0)
+    massCoreNormal     =+(                          & ! Equation (15) of Chan et al. (2022; MNRAS; 551; 943; https://ui.adsabs.harvard.edu/abs/2022MNRAS.511..943C).
```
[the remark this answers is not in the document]
yzhaoastro: I double-checked the reformatted equation and it looks good.

source/dark_matter_profiles_DMO.soliton_NFW.F90
```diff
@@ -345,20 +430,49 @@ subroutine solitonNFWComputeProperties(self,node,radiusVirial,radiusScale,radius
...
+         &                  rangeExpandUpwardSignExpect  =rangeExpandSignExpectNegative, &
+         &                  rangeExpandType              =rangeExpandMultiplicative      &
          &                 )
-    radiusCore_  =radiusCore
-    radiusScale_ =radiusScale
-    densityScale_=densityScale
-    densityCore_ =densityCore
-    radiusSoliton=finder%find(rootGuess=3.0d0*radiusCore)
+    radiusSoliton=finder%find(rootGuess=3.0d0*radiusCore, status=status)
+    if (status /= errorStatusSuccess) then
```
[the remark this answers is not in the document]
yzhaoastro: I've updated the code so that the first calculation is now part of the loop, with an if to handle the first iteration differently by retrieving the stored random offset if available, and sampling a new one for subsequent iterations.

PR #932, opened 2025-09-27
Use `readOnly=.true.` when reading excursion set files

abensonca:
This avoids lock conflicts in the HDF5 library.


abensonca:
Closing - wrong approach.

PR #979, opened 2025-12-04
Update FDM core–halo relation and add validation test

yzhaoastro:
This PR updates the FDM mass–core relation to ensure consistency with Chan et al. (2022). The red curve is computed directly from Eq. (15) of Chan et al. (2022), confirming our cosmological and scaling inputs.
The blue curve is the scatter-free core mass from the node-operator integration, and the green curve includes the ±50% parameter uncertainty.
All three lie within the purple shaded region of Figure 5 in Chan et al. (2022), matching the expected core–halo relation.
image
Updates:
Corrected the soliton mass–core relation (integration and analytic forms now consistent with Chan+22).
Renamed the scatter-free integrated core mass to massCoreNormal.
Added a new scatter-aware core mass, massCore, which includes ±50% parameter uncertainty and allows computing radiusSoliton.
Added a new validation test: validate-fuzzyDarkMatter.
Updated the FDM reference parameter file accordingly.

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -944,7 +944,7 @@ jobs:
       artifact: galacticus-exec
       runPath: ./testSuite
       cacheData: 1
-      ompNumThreads: 3
+      ompNumThreads: 2
       uploadFile: ./testSuite/outputs/*.json
       uploadName: validate-fuzzyDarkMatter
     secrets: inherit


abensonca (approved):
This all looks good. I made some minor changes to clean up formatting and two more significant changes:

1. I changed the names of the meta-properties (and the names that are used for these in the output) to have a `soliton` prefix. This should make it more obvious what they refer to in the output, and should avoid conflicts in the code itself (for example, some other model in the code might want to add a `massCore` meta-property, which would conflict with ours, but renaming ours to `solitonMassCore` should help avoid this).

2. Some of the `change`s in `fuzzyDarkMatter.xml` were just introducing `<darkMatterProfileDMO value="NFW"/>` into some of the sections. Instead of replacing the entire sections, I modified it to just insert `<darkMatterProfileDMO value="NFW"/>`  where needed.

I ran the `validate-fuzzyDarkMatter.py` test, and it completed successfully, so I think I did this correctly. But you should look through and check that the changes I made seem reasonable.

If you're happy with the changes, I'll merge this PR once the tests pass.

PR #992, opened 2026-01-12
Add solitonNFW heating model

yzhaoastro:
This PR adds a solitonNFWHeated dark matter profile and two soliton-specific node operators.

The solitonNFWHeated model applies the standard CDM tidal heating/stripping to the NFW-like envelope, while evolving the solitonic core using the prescription in [Du et al. (2018)](https://journals.aps.org/prd/abstract/10.1103/PhysRevD.97.063507).

Two node operators are included:
(1) tidal heating/stripping of soliton cores following [Du et al. (2018)](https://journals.aps.org/prd/abstract/10.1103/PhysRevD.97.063507), and
(2) a core–core merger rule for soliton cores based on [Schwabe et al. (2016)](https://journals.aps.org/prd/abstract/10.1103/PhysRevD.94.043513).

### DIFF
--- a/source/nodes.operators.physics.satellite_merging.soliton.F90
+++ b/source/nodes.operators.physics.satellite_merging.soliton.F90
@@ -326,8 +326,10 @@ contains
          &                                              )
     call darkMatterProfileHost%floatRank0MetaPropertySet(                                                                    &
          &                                                self_%randomOffsetID                                             , &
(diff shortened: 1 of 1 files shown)


abensonca (approved):
I made some changes to clean up the formatting and remove some empty functions. Everything else looks good - I will commit this once all tests have passed.

PR #1004, opened 2026-02-27
Allow initializing satellite’s bound mass in a more flexible way

Xiaolong-Du:
In the current version, the satellite's bound mass can be initialized based on a specified density contrast through the `densityContrastMassBound` parameter. This density contrast is defined relative to the matter density. If one needs to define the bound mass using the $200\rho_{\rm crit}$ definition, `densityContrastMassBound` will be redshift-dependent. With this update, one can easily do this through a class object `virialDensityContrastDefinition`, e.g.

```
 <componentSatellite value="orbiting" >
    <initializationTypeMassBound value="densityContrast"/>
    <virialDensityContrastDefinition value="fixed">
      <densityType value="critical"/>
      <densityContrastValue value="200.0"/>
      <turnAroundOverVirialRadius value="2.0"/>
    </virialDensityContrastDefinition>
  </componentSatellite>
```

### DIFF
--- a/source/objects.nodes.components.satellite.orbiting.F90
+++ b/source/objects.nodes.components.satellite.orbiting.F90
@@ -120,19 +120,19 @@ module Node_Component_Satellite_Orbiting
   !!]
 
   ! Objects used by this module.
-  class(virialDensityContrastClass), pointer :: virialDensityContrast_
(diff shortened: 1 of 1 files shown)


abensonca (approved):
This all looks good, and all tests have passed. I'll merge this into `master`. Thanks for the PR!

PR #1005, opened 2026-02-27
Fix a bug related to drawing the coordinates of the N-body particles

Xiaolong-Du:
There is a bug in drawing the coordinates of N-body particles. In the current implementation, the radial coordinate $r$ of each particle is drawn uniformly from the interval $[0,R_{\rm truncate}]$:
```
call positionSpherical%rSet(radiusTruncate_*randomDeviates(3))
```
This leads to a density profile of $r^{-2}$ regardless of the actual halo profile. To fix this issue, one should first draw the enclosed mass $M_{\rm enclosed}$ uniformly from the interval $[0,M_{\rm truncate}]$. Then compute the radius enclosing $M_{\rm enclosed}$.

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -1396,7 +1396,7 @@ jobs:
                 testSuite/regressions/linearGrowthForFutureModel.xml,
                 testSuite/regressions/ludlowScaleSetBeforeTreeInitialization.xml,
                 testSuite/regressions/mergerAtFinalTimeInTree.xml,
-                testSuite/regressions/particulate.xml,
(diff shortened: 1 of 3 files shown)


abensonca:
I pushed  a fix for the problem. I also extended the particulate test in the test suite so that it verifies that the mass profile of the particles is consistent with Galacticus' internal expectations for the mass profile. In this test we have to account for Poisson fluctuations in the number of particles when testing agreement. If currently allow up to 3σ deviation before marking failure. It's possible we'll need to increase this threshold if we get occasional >3σ fluctuations.

PR #1140, opened 2026-06-03
fix(starFormation): perform adaptive SFH table writes only under an exclusive file lock

abensonca:
## Summary

Fixes a file-locking race in `adaptiveConstructorInternal` (`source/star_formation.histories.adaptive.F90`), the same class of bug fixed for `mass_distributions.Gaussian_ellipsoid.F90` in 249bd0ed1.

The routine checked `File_Exists` **outside** any file lock, then branched into either a shared-lock read or an exclusive-lock build/write of the HDF5 table of adaptive star formation history time bins. Because the existence test was unlocked, the check-then-act was not atomic:

- two threads could both observe the file as missing and both enter the build/write path, or
- a reader could take a shared lock on a file that another thread was concurrently creating/writing,

occasionally corrupting the file.

## Fix

Wrap the read/build logic in a two-pass loop mirroring the Gaussian-ellipsoid fix:

- **Pass 1 — shared lock:** if the file exists (checked *under* the lock), read it and exit; otherwise release the shared lock and retry.
- **Pass 2 — exclusive lock:** re-check existence under the exclusive lock; read if another thread won the race, otherwise build and write.

Writes therefore only ever occur while an exclusive lock is held, while still avoiding redundant table builds. `Directory_Make` is hoisted ahead of the lock loop because the lock file (`<fileName>.lock`) is created alongside the data file and the directory must exist before the first `File_Lock` on either pass. The write retains `overWrite=.false.`, which is consistent with pass 2 having confirmed the file does not exist under the exclusive lock.

## Testing

- Compiles cleanly.
- Pre-commit hooks (Fortran static analysis, XML/LaTeX fragments, etc.) pass.


abensonca:
:warning: This will almost certainly fail since the build environment has already been updated for the incoming GCC 16 changes. Once those are merged, just merge them into this branch and re-run CI/CD. :warning:

PR #1154, opened 2026-06-06
fix: Restore refClass documentation anchors for functionClass implementations

abensonca:
## Problem

The link checker began reporting many broken `\refClass{…}` links after the recent large merge into `master` — e.g. `class.posteriorSampleLikelihoodIndpndntLklhdsSqntl`, `class.virialOrbitJiang2014`, and every other functionClass *implementation* class. The link checker was correct: these anchors genuinely no longer existed in the published `Galacticus_Development.pdf`.

## Root cause

The Perl→Python port of `ClassDocumentation` added a filter (`_is_inserted_from_other_source_file`) that skipped `type` nodes whose `source` differed from the tree being processed, assuming each child class file's own `preprocess.py` pass would document its type.

That assumption is false. FunctionClass implementations compile as **submodules** (`83f62a184a`): a child file's standalone pass yields a `submodule (...) <name>_` containing only `module procedure` bodies — the `type, extends(...) :: <name>` definition lives **only** in the parent functionClass module, inserted there by FunctionClass carrying the *child's* source filename. The filter therefore dropped every implementation type from its `classes.xml`, removing its `\hyperdef{class}{<name>}` anchor from `dataMethods.tex` / the Development PDF and breaking every `\refClass{<name>}` link to it.

The Perl original never filtered on `source`, which is why the links worked before.

## Fix

- Remove the `_is_inserted_from_other_source_file` filter, restoring the Perl original's behaviour of recording every `type` node. (The `is_inner_process_tree_call()` guard against double-processing generic clones is untouched.)
- Invert the test that encoded the buggy behaviour (`…skips_types_from_other_source_files` → `…records_types_from_other_source_files`).
- Fix three pre-existing `\refClass{outputAnalysisTargetData}` typos (a functionClass directive name, which has no class anchor) in `output.analyses.target_data.standard.F90`, repointing them at the documented base type `outputAnalysisTargetDataClass`. (Unrelated to the porting bug; introduced 2026-05-07.)

## Verification

Ran a documentation build (preprocess → `extractData.py` → Development PDF) and validated against the rebuilt PDF's named destinations (exactly what the link checker inspects):

| Metric | Before | After |
|---|---|---|
| `class.*` named destinations in `Galacticus_Development.pdf` | 471 | **2042** |
| Broken `refClass` links (4143 references, 1288 distinct targets) | many | **0** |

All 24 `test_class_documentation.py` tests pass.

### DIFF
--- a/python/Galacticus/Build/SourceTree/Process/ClassDocumentation.py
+++ b/python/Galacticus/Build/SourceTree/Process/ClassDocumentation.py
@@ -20,7 +20,7 @@ from List.ExtraUtils                     import as_array
 from XML.Utils                           import xml_to_dict
 from Fortran.Utils                       import UNIT_OPENERS
 from Galacticus.Build.Directives         import extract_directives
-from Galacticus.Build.SourceTree         import walk_tree, parse_file
+from Galacticus.Build.SourceTree         import walk_tree, parse_file, parse_code
 from Galacticus.Build.SourceTree.Parse.Declarations import parse_declaration
 from Galacticus.Build.SourceTree.Process import (
     register_postprocess, is_inner_process_tree_call,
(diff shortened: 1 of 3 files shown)


abensonca:
Follow-up commit (`fe86f0ec72`): the documentation build surfaced a second, related regression from the submodule split.

**Symptom:** `Galacticus_Development.tex` failed to compile — `doc/dataMethods.tex` contained empty `\begin{description}...\end{description}` environments (LaTeX: *"Something's wrong--perhaps a missing \item"*). 78 such empty blocks, mostly `enumeration*Type` classes plus iterator/table/container helper types.

**Root cause:** auxiliary helper types defined alongside a functionClass implementation (e.g. `domainIteratorCartesian3D` in a `<computationalDomain>` file) whose `<methods>` directive describes a function-typed method got no `<type>` in `*.classes.xml`. Since functionClass implementations now compile as submodules (83f62a184a), `FunctionClass` inserts each contained procedure's *module interface* into the parent tree as raw `code` blobs rather than parsed function nodes — so `walk_tree` never yielded them and `_process_function` couldn't attach a return type. The type-less method then produced an empty `description` environment.

**Fix:** after the normal type-resolution pass, if any described method has a resolved binding but still no type, re-parse just that procedure's small `interface … end interface` blob and resolve from it. Parsing the small per-procedure blobs individually (and only those naming an unresolved function) avoids `parse_code`'s super-linear cost — a whole-tree re-parse of the biggest assembled modules (`output.analyses`, `nodes.operators`) cost many minutes, whereas the targeted approach handles them in seconds (7.6 s / 13.6 s).

**Verification:** regenerated all `classes.xml`, rebuilt `dataMethods.tex` → **0 empty description blocks (was 78)**; the Development manual now builds (**1330 pages**) with all **2043 `class.*`** named destinations present (incl. `domainIteratorCartesian3D`, `enumerationmassType`, `posteriorSampleLikelihoodIndpndntLklhdsSqntl`). All 105 SourceTree unit tests pass, including a new regression test for the interface-blob recovery.

PR #1166, opened 2026-06-14
fix(objectBuilder): guard against infinite recursion during object building (#397)

### LINKED ISSUE #397: Avoid infinite recursion during object building from the parameter file
Objects which composite a member of their own class can get into an infinite recursion if no such object is explicitly provided to them (they instead try to composite themselves). Add an option to the `objectBuilder` directive (or maybe even automate this) that indicates that an object of the class must be explicitly provided to avoid this problem.

abensonca:
Closes #397.

## Problem
An object that composites a member of **its own class** — directly (e.g. `galacticFilterNot`, which wraps another `galacticFilter`) or via another, **mutually-compositing** class (e.g. `criticalOverdensityEnvironmental` ↔ `haloEnvironmentNormal`) — could, if no such object was provided explicitly, search up the parameter tree, re-discover the object currently being built, and try to build it again → **unbounded recursion → stack overflow (SIGSEGV) with no diagnostic**.

Notes from the investigation (see #397 comments):
- The pre-existing same-class `objectBuilder` guard only caught *direct* self-composition and emitted a cryptic message.
- Cross-class mutual recursion is reachable whenever the two classes share a module (so the circular-`use` compile barrier doesn't apply). `criticalOverdensity`/`haloEnvironment`/`cosmologicalMassVariance` all live in `cosmological_density_field` and cross-compose — this was reproduced as a real SIGSEGV.
- The `functionGlobal` wrappers (`nodeOperatorConstruct`, `accretionDiskSpectraConstruct`) can subvert the module hierarchy, but they delegate to the same generated factory, so a factory-level guard covers them too.

## Fix (Option C)
- **Factory-level build-stack guard (general):** a thread-private build stack in `Input_Parameters` (`Input_Parameters_Build_Stack_Push`/`_Pop`). Every generated functionClass factory pushes the parameter node it is about to build and pops it afterwards, aborting with an informative error **and the build stack** if the same node/class is already being built. Because both the ordinary `objectBuilder` path and the `functionGlobal` wrappers route through the factory, this catches self-composition, cross-class mutual recursion, and functionGlobal-mediated recursion alike. The push/pop sit **after** the `recursive="yes"` short-circuit, so legitimately self-recursive classes are unaffected.
- **Clearer same-class message:** the existing `objectBuilder` guard now tells the user the object must be provided explicitly.

## Behaviour
- Empty `<galacticFilter value="not"/>` → `a [galacticFilter] composites a member of its own class … provide a [galacticFilter] explicitly`.
- Environmental crit ↔ normal halo environment → `recursive build of [criticalOverdensity] detected …` + `Build stack (outermost first): -> [criticalOverdensity] -> [haloEnvironment] -> [criticalOverdensity] <-- recursion`.
- Both abort cleanly (exit 1) instead of SIGSEGV; valid configurations are unaffected.

## Tests
- `python/Galacticus/Build/SourceTree/tests/test_object_builder_recursion_guard.py` — pins both guards in the codegen (fast pytest job).
- `testSuite/test-objectBuilder-recursion.py` (+ `recursionGuardSelfComposition.xml`, `recursionGuardMutual.xml`) — runs both recursion modes and asserts a clean abort (not a segfault); added to the `Miscellaneous` matrix in `cicd.yml`.

## Verification (local, full rebuild)
- Guard present in 192 factories; both recursion modes abort cleanly (exit 1); valid nested-filter model completes (exit 0); full preprocessor pytest suite **421 passed**.

## Note for reviewers
The Makefile regenerates `*.p.F90` only when the **source** `.F90` changes, not when the `FunctionClass`/`ObjectBuilder` codegen changes — so after pulling onto an existing `work/build`, force regeneration with `touch source/*.F90 && make` (a clean checkout is unaffected). Pre-existing limitation; happy to address as a follow-up if desired.

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -1419,6 +1419,7 @@ jobs:
                 test-postprocessing.py,
                 test-merger-tree-export.sh,
                 test-AGNSED.py,
+                test-objectBuilder-recursion.py,
                 test-library-interfaces.py ]
     uses: ./.github/workflows/testModel.yml
     with:
--- a/python/Galacticus/Build/SourceTree/Process/FunctionClass/__init__.py
+++ b/python/Galacticus/Build/SourceTree/Process/FunctionClass/__init__.py
@@ -3118,7 +3118,7 @@ def _generate_constructor(directive, classes_ordered, non_abstract_classes,
             )
         post['content'] += (
             '        call Input_Parameters_Build_Stack_Push(subParameters%parameters,'
-            f"'{directive_name}',{loc_expr})\n"
+            f"'{directive_name}',{'.true.' if allow_recursion else '.false.'},{loc_expr})\n"
             '        select type (self)\n'
             f'          type is ({target_name})\n'
             f'            self={target_name}(subParameters)\n'
@@ -3169,7 +3169,7 @@ def _generate_constructor(directive, classes_ordered, non_abstract_classes,
         '      subParameters=parameters%subParameters'
         '(char(parameterName_),copyInstance=copyInstance_)\n'
         '      call Input_Parameters_Build_Stack_Push(subParameters%parameters,'
-        f"'{directive_name}',{loc_expr})\n"
+        f"'{directive_name}',{'.true.' if allow_recursion else '.false.'},{loc_expr})\n"
         '      select case (char(instanceName))\n'
     )
     for c in non_abstract_classes:
--- a/python/Galacticus/Build/SourceTree/Process/FunctionClass/__init__.py
+++ b/python/Galacticus/Build/SourceTree/Process/FunctionClass/__init__.py
@@ -3040,7 +3040,8 @@ def _generate_constructor(directive, classes_ordered, non_abstract_classes,
     post['content'] += '      !!}\n'
     post['content'] += (
         '      use :: Input_Parameters  , only : inputParameter         , '
-        'inputParameters\n'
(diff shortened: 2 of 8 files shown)


abensonca:
Follow-up: fixed a false positive in the build-stack guard (commit b0b055568).

**Symptom:** `subhaloOrbitOnZAxis.xml` aborted with `recursive build of [geometryLightcone] detected`.

**Why it was wrong:** that `geometryLightcone` cycle passes through `starFormationHistoryFixedAges`, which is `recursive="yes"`. Its factory short-circuits the re-entry (returns a reference to the object already under construction), so the cycle is **bounded** — `geometryLightcone` is built exactly twice, not infinitely. The guard fired on the repeat before the short-circuit could break the cycle.

**Fix:** a repeated `(node, class)` is only flagged when **no recursion-aware (`recursive="yes"`) class lies between** the earlier build and the current one on the stack. The factory threads its `allow_recursion` flag into `Input_Parameters_Build_Stack_Push`; each stack entry records it, and the cycle check scans the intervening entries, suppressing the error when such a class is present.

**Verified (full rebuild):**
- `subhaloOrbitOnZAxis.xml` → runs to completion (exit 0). It's already in the CI model matrix, so it guards against this regression.
- Self-composition (`galacticFilter not`) and mutual recursion (`criticalOverdensity` ↔ `haloEnvironment`) → still detected and reported cleanly.
- A valid environmental/normal model → evolves normally (no false positive).
- Preprocessor pytest suite → 421 passed (the codegen test now also pins the recursion-aware flag).

PR #1207, opened 2026-06-29
feat: add Python parameter-file resolver (XInclude + change files)

abensonca:
## Summary

First stage of the Python parameter-file **resolver/preprocessor** (revised A4 / Pillar B). Adds `Galacticus.Parameters.resolve`, which re-implements on an lxml tree the two file-level transformations Galacticus bakes eagerly into the DOM, so a fully-resolved file can later be handed to `Galacticus.exe`:

- **XInclude** — Galacticus' limited `xpointer(parameters/*)` form, strict file-relative `href` resolution (matching `source/utility/IO/XML.F90`), recursive, with `<xi:fallback>`.
- **Change files** — the full operation set (`remove`, `update`[+`append="true"`], `append`, `insertBefore`, `insertAfter`, `replace`, `replaceOrAppend`, `replaceWith`, `encapsulate`), applied in command-line order, each `<change>` in document order, matching `source/utility/input_parameters.F90`.

A shared path helper gives Galacticus' legacy "bare path is absolute from the `<parameters>` root" semantics while letting **full XPath** (predicates, `//`, axes, functions — via lxml) pass through.

## Deliberately *not* done here (see design notes)

- **`id`/`idRef` are not dereferenced.** They are runtime *pointers* to a single shared object; inlining would make Galacticus build separate copies, changing behaviour. Reference validation + conditionals are a later stage.
- **`=[...]` math expressions are left intact** for Galacticus to evaluate at runtime (deferred until a fuller front end, since they must be re-evaluated per MCMC step).

## Verification

- `python/Galacticus/Parameters/tests/test_resolve.py` — 22 unit tests (path resolution, XInclude, every change operation, ordering, error cases). Runs in the normal `pytest` stage; no binary needed.
- `testSuite/test-parameter-resolver.py` — **differential test against the Galacticus oracle** (`Galacticus.exe --output-processed-parameters`, which serializes the processed DOM before any value is read, so it faithfully reflects XInclude + changes). Every change operation and **every XInclude-using parameter file in the corpus (26/26)** match the binary's output exactly. Skips cleanly if no built binary is present.

This is a pure library + tests addition — no wiring into the `galacticus` launcher yet (that's a later stage), so nothing in existing behaviour changes.

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -43,10 +43,10 @@ jobs:
         with:
           python-version: '3.x'
       - name: Install pytest
-        # platformdirs + requests are needed by the galacticus_launcher tests
-        # (the only third-party deps in the python/ pytest suite); both are tiny
-        # pure-Python packages so the job stays fast.
-        run: python -m pip install --upgrade pip pytest platformdirs requests
+        # platformdirs + requests are needed by the galacticus_launcher tests and
+        # lxml by the parameter-resolver tests (the third-party deps in the
+        # python/ pytest suite); all are small wheels so the job stays fast.
+        run: python -m pip install --upgrade pip pytest platformdirs requests lxml
       - name: Run Python regression tests
         run: |
           cd $GITHUB_WORKSPACE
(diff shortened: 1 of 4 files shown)


abensonca:
Part of #1208 (tracking issue for the parameter-file resolver/preprocessor; this PR is Stages 0–1).

Issue #94, opened 2020-03-09
Use generics to reduce duplicated code in the IO HDF5 module


abensonca:
Read/write functions for different types and ranks should be collapsible into single generic functions.

abensonca:
Depends on #92

abensonca:
For many of the functions we don't need generics - using assumed rank arrays for the arguments would let a single function handle dataset writes (for example) for all possible ranks.

abensonca:
# Analysis: recommended approach

**Recommendation: a hybrid — the `<generic>` preprocessor template over the *type* axis, and Fortran 2018 assumed-rank (`dimension(..)`) over the *rank* axis.** This is exactly the pattern already proven in `source/utility/polyranks.F90`, and neither mechanism alone captures the full win.

## Duplication structure

`source/utility/IO/HDF5/_module.F90` is 18,340 lines (the largest file in the repo) with ~90 near-duplicate procedures spanning {integer, integer8, double, character, varying_string, var-length} × {rank 1–6} × {allocatable/static} × {read/write} × {dataset/attribute}. Comparing variants:

- **Across ranks**, only 3–15 lines differ per ~170-line procedure: the `dimension(:,:,...)` declarations, the `datasetRank` literal, the `datasetDimensions` extent, and the append-dimension filtering (trivial for 1D). The `h5dwrite_f`/`h5dread_f` calls themselves are **identical across all ranks**.
- **Across types**: the `H5T_*` constant, the `c_loc` + contiguous-copy machinery for `kind_int8`, and `h5tcopy`/`h5tset_size` datatype setup for character.
- **Allocatable vs. static reads** differ by ~5 lines out of ~370 (allocate vs. bounds check). There is already copy-paste drift as evidence of the maintenance cost — e.g. a duplicated open-check block at lines 9598 and 9603.

## Why each axis gets a different tool

[...]

Issue #92, opened 2020-03-09
Use Fortran 2003 API for HDF5 read/writes


abensonca:
Then we don't need to go through C-API for region references for example, and might also make reading of different rank/types more similar so easier to implement as generics via preprocessor.

abensonca:
# Report: Retiring HDF5 C-API wrappers in favour of the Fortran 2003 API

## 1. What exists today

All HDF5 output flows through `source/utility.IO.HDF5.F90` (~18,400 lines). It uses the HDF5 **Fortran** API for the overwhelming majority of operations, but falls back to the **C API** (and three bespoke C helpers) for a handful of datatypes that the *legacy, typed* Fortran interfaces could not express. These fall-backs are declared in one `interface` block (`utility.IO.HDF5.F90:388-460`) plus one in `error.F90:52-55`, and are backed by three small C files:

| C symbol | Source | Kind | Call sites |
|---|---|---|---|
| `H5Dread` | HDF5 C API | dataset read (c_ptr buffer) | 36 (27 region-ref, 5 int8, 4 VLEN) |
| `H5Dwrite` | HDF5 C API | dataset write (c_ptr buffer) | 4 (VLEN) |
| `H5Aread` | HDF5 C API | attribute read (c_ptr buffer) | 4 (3 int8, 1 var-string) |
| `H5Awrite` | HDF5 C API | attribute write (c_ptr buffer) | 3 (2 int8, 1 compound) |
| `H5TBread_fields_name` | HDF5 HL C API | read one table column | 1 (int8 column) |
| `H5T_C_S1_Get` | `hdf5_cTypes.c` | return `H5T_C_S1` | 3 |
| `H5T_Variable_Get` | `hdf5_cTypes.c` | return `H5T_VARIABLE` | 2 |
| `HDF5_Unit_Type_Create` | `hdf5_cTypes.c` | build `unitType` compound type | 1 |
| `H5Close_C` | `hdf5_cFuncs.c` | call `H5close()` | 1 (`error.F90:792`) |

[...]

PR #1210, opened 2026-07-02
test: extend HDF5 I/O unit tests to cover all read/write variants

abensonca:
## Purpose

First step of the campaign to deduplicate the HDF5 I/O module (#94, folding in the remainder of #92, per the plan in https://github.com/galacticusorg/galacticus/issues/94#issuecomment-4861197632): establish a complete test safety net covering **every** read/write variant before any are collapsed into generic implementations.

## Coverage added

Round-trip tests for all previously-untested variants:

- **integer 2-D and 3-D datasets**
- **long integer 2-D, 3-D, and 4-D datasets** — with values exceeding the 32-bit range to confirm genuine 64-bit storage
- **double 6-D datasets** — including extensible datasets, appending to an arbitrary dimension, and subsection reads (the append/hyperslab logic is the most rank-dependent code in the module)
- **2-D arrays of variable-length double rows** (`VarDouble_2D`, the one untested VLEN variant)
- **2-D double attributes** and **logical attributes**
- **table column reads** — real, integer, long integer, and character columns, plus a subsection read of the long integer column, which exercises the one remaining `H5TBread_fields_name` C-API wrapper. The table is generated with h5py as a compound-type dataset following the H5TB conventions.

Variants with no corresponding Fortran reader (2-D double attribute, 3-D integer dataset, 4-D long integer dataset) are verified externally via a new h5py verifier script (`verify_attributes.py`), following the existing `verify_units.py` pattern.

The group dataset-listing test is updated for the new datasets (20 → 27).

## Notes

- Region-reference reads, VLEN varDouble/varInteger8, int8 attributes/datasets (1-D), and the units compound attribute were already covered by the existing tests.
- Verified locally: all 206 assertions pass in both the chunking-enabled and chunking-disabled passes.

Part of #94; supports the remaining work on #92.

### DIFF
--- a/source/tests/IO/HDF5.F90
+++ b/source/tests/IO/HDF5.F90
@@ -36,7 +36,7 @@ program Tests_IO_HDF5
        &                                                                       integerValueReread             , i                    , &
        &                                                                       j                              , k
   logical                                                                   :: appendableOK
-  integer                                                                   :: unitsStatus
+  integer                                                                   :: unitsStatus                    , attributesStatus
   type            (unitType       )                                         :: unitsValue
   integer                                       , dimension(10)             :: integerValueArray
   integer                                       , dimension(10)             :: integerValueArrayRereadStatic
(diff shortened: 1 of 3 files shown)


abensonca (commented):
Reviewed - all looks good.

PR #1211, opened 2026-07-02
refactor: collapse HDF5 attribute read/write functions using generics

abensonca:
## Purpose

Second step of #94: collapse the numeric attribute read/write families in the HDF5 I/O module into templated generic implementations. This is the proof-of-pattern PR for the hybrid design (preprocessor `<generic>` over the *type* axis, Fortran 2018 assumed-rank over the *rank* axis) that the dataset families will follow.

Stacked on #1210 (the test safety net) — merge that first; this PR's base is `feature/hdf5-generics-tests`.

## What changed

**16 hand-written procedures → 4 templates** (net **−1,077 lines**):

| Family | Before | After |
|---|---|---|
| `writeAttribute` (numeric) | 7 procedures (int/int8/double × scalar/1-D[/2-D]) | 1 template, assumed-rank `dimension(..)` |
| `readAttribute` (numeric) | 6 procedures | 2 templates (scalar, 1-D allocatable) |
| `readAttributeStatic` (numeric) | 3 procedures | 1 template |

- The assumed-rank write extends capability: attributes of **any rank up to 7** are now supported for all three templated types (previously only double had a 2-D variant).
- Reads stay rank-specific — they're constrained by allocatability (an assumed-rank `allocatable` dummy would force allocatable actuals at every scalar call site) and by the pseudo-scalar convention — but collapse over the type axis.
- Character, varString, logical, and units variants are untouched (they don't fit the numeric template).
- Call sites are unaffected: the `writeAttribute`/`readAttribute`/`readAttributeStatic` generic bindings are unchanged.

## Unified calling convention (load-bearing finding)

All templated variants now use `c_loc()` of a `target, contiguous` buffer with the `c_ptr` variants of `h5awrite_f`/`h5aread_f` — the convention int8 already used. Two important facts established empirically (gfortran 16 + HDF5 1.14.5):

1. **The HDF5 1.14 Fortran API has no assumed-type/assumed-rank buffer interfaces.** The typed generics are kind×rank-specific (generated `H5_gen` module, ranks 0–7), so an assumed-rank buffer *cannot* be passed to them directly. The `c_ptr` convention is what makes the assumed-rank template possible. This corrects a premise in the issue-#94 plan; the same applies to the upcoming dataset-family PRs.
2. `c_loc()` on a `contiguous` assumed-rank dummy is valid (TS29113) and the `contiguous` attribute makes the compiler handle copy-in/out — eliminating the manual "allocate a contiguous copy" workarounds the int8 paths carried.

## Also in this PR

- **`Allocate.py` fix**: the `<allocate>` directive preprocessor crashed on declarations whose *type* is a generic placeholder (`{Type¦intrinsic} :: x`) since these don't parse as declarations pre-expansion; it now defers to the post-generics rerun (as it already did for placeholder *attributes*).
- **Latent bug fix**: the integer scalar read's error messages had a doubled `locationReport()` (copy-paste drift) — normalized.

## Verification

- All **206 assertions** in `tests.IO.HDF5.exe` pass (the #1210 safety net covers every templated variant, including the int8 cross-type reads and pseudo-scalar paths).
- Full `Galacticus.exe` build succeeds — all call sites across the codebase resolve against the new generics.
- `parameters/quickTest.xml` runs end-to-end; output attributes spot-checked via h5py (int64/float64/string attributes land with correct on-disk types).

Part of #94.

### DIFF
--- a/python/Galacticus/Build/SourceTree/Process/Allocate.py
+++ b/python/Galacticus/Build/SourceTree/Process/Allocate.py
@@ -45,6 +45,23 @@ def _declaration_has_generic_placeholder(parent, variable):
     return any(isinstance(f, str) and '¦' in f for f in fields)
 
 
+def _inside_generic_template(node):
+    """Return True if `node` is inside a construct whose opener carries an
+    unresolved generic-template placeholder (e.g. a subroutine named
+    `Foo{Type¦label}`) — in which case declarations inside it may not yet be
+    parseable at all (a declaration whose *type* is a placeholder, such as
+    `{Type¦intrinsic} :: x`, is not recognized as a declaration until the
+    placeholder is substituted).
+    """
+    p = node.get('parent')
+    while p is not None:
+        opener = p.get('opener')
+        if opener and '¦' in opener:
+            return True
+        p = p.get('parent')
+    return False
+
+
 def process_allocate(tree, options):
     """Mirrors Process_Allocate() from Allocate.pm."""
     for node in walk_tree(tree):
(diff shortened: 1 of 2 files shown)


abensonca (commented):
Reviewed - all looks good.

PR #1212, opened 2026-07-02
refactor: collapse HDF5 numeric dataset write functions using generics

abensonca:
## Purpose

Third step of #94: collapse the numeric `writeDataset` family — the thickest concentration of rank-dependent code in the module — into a single templated implementation. Stacked on #1211 (attribute family), which is stacked on #1210 (test safety net); merge in that order.

## What changed

**13 hand-written procedures → 1 template** (net **−2,088 lines**): integer 1–3D, long integer 1–4D, and double 1–6D dataset writes become one assumed-rank (`dimension(..)`) procedure templated over {integer, integer8, double} via the `<generic>` directive introduced in #1211.

Key points:

- **The rank-generic append logic subsumes the 1-D special case exactly.** With `appendDimensionActual=1`, the filtered-dimensions consistency check passes trivially for rank 1, so the previously separate 1-D append path (no `appendDimension` argument, simpler hyperslab arithmetic) falls out of the general code with identical behavior.
- **Capability gains:** `appendDimension` is now accepted for all templated types and ranks (previously only 2-D+ double and 2-D int8), with a new bounds check; ranks up to 7 are supported for all three types (previously integer stopped at 3-D, int8 at 4-D, double at 6-D).
- **Unified calling convention** (same as #1211): `c_loc()` on a `target, contiguous` buffer + the `c_ptr` variant of `h5dwrite_f`. The int8 "allocate a contiguous copy" workarounds are gone — the `contiguous` attribute makes the compiler handle copy-in/out.
- **API normalizations:** the int8 4-D variant's `commentText` argument is renamed to `comment` (verified no keyword-form callers exist); dataset rank is validated explicitly (assumed-rank would otherwise accept scalars at compile time).
- Character, varString, and the VLEN variants are untouched.
- Call sites are unaffected; the `writeDataset` generic binding is unchanged.

## New regression test

Zero-sized dataset writes previously went through the typed HDF5 interface for integer/double and now go through `c_loc` — a new test writes and re-reads zero-sized double and integer 1-D datasets (unchunked, since chunked datasets require chunk extents ≥ 1) to pin this behavior.

## Verification

- All **210 assertions** in `tests.IO.HDF5.exe` pass, including the extensible/append-to-arbitrary-dimension round-trips for every rank 1–6, the integer 2–3D / int8 2–4D coverage added in #1210, and the h5py-verified write-only datasets.
- Full `Galacticus.exe` build succeeds — all call sites resolve.
- `parameters/quickTest.xml` runs end-to-end; output verified via h5py (36 nodeData datasets, float64 and int64, correct on-disk types and values).

Part of #94.


abensonca (commented):
Reviewed - all looks good.

PR #1213, opened 2026-07-02
refactor: collapse HDF5 numeric dataset read functions using generics

abensonca:
## Purpose

Fourth step of #94: collapse the numeric `readDataset`/`readDatasetStatic` families — the largest tranche, carrying the region-reference and subsection/selection machinery. Stacked on #1212 → #1211 → #1210; merge in that order.

## What changed

**21 hand-written procedures → 2 templates** (net **−6,041 lines**): integer 1–2D, long integer 1–3D, and double 1–6D reads (allocatable and static) become two assumed-rank procedures templated over {integer, integer8, double}. The module now stands at **9,106 lines, down from 18,340** at the start of the campaign (−50%).

New rank-generic machinery (each proven in a standalone experiment before use):

- **`select rank` allocation**: the allocatable variant allocates its assumed-rank `intent(out)` dummy inside `rank (n)` blocks, where the associating entity retains the `allocatable` attribute (the `polyranks.F90` pattern).
- **Runtime-allocated dimension arrays**: `rank()` of an `intent(out)` dummy is not permitted in specification expressions, so the dimension work arrays are allocatable locals sized at runtime.
- **Rank-generic selection map**: the `readSelection` element map is built by mixed-radix enumeration of the leading dimensions, reproducing the original 1-D/2-D/3-D point orderings exactly. `readBegin`/`readCount`/`readSelection` are now uniformly available for all templated types and ranks 1–7 (previously selection reads existed only for int8 ranks 1–3 and double ranks 1–2), with new run-time validation that subsection array sizes match the dataset rank.
- All reads use the unified `c_loc()` + `c_ptr` convention; the int8 contiguous-copy workarounds are gone.

## Latent defects fixed

- The double 1-D allocatable variant had a duplicated open-object check (copy-paste drift).
- The 1-D allocatable `readSelection` path created its memory dataspace from `shape()` of the **not-yet-allocated** output array (`intent(out)` allocatables arrive deallocated); all other ranks correctly used the selection dimensions.

## The safety net earned its keep

My first cut of the static template used the allocatable variant's contiguous memory-space layout for referenced reads. The #1210 test suite immediately failed 8 assertions — referenced regions read into static arrays must land in the caller's buffer *corner* (memory space shaped like the full buffer, hyperslab-selected), not contiguously — a layout difference invisible at rank 1 and silently corrupting at rank ≥ 2. Exactly the class of regression these tests were built to catch.

## Verification

- All **210 assertions** pass, including referenced reads (ranks 1–5, static/allocatable/partial), subsection reads for every rank, and cross-type reads (int→int8, u32→int8).
- Full `Galacticus.exe` build succeeds — all call sites resolve.
- `parameters/quickTest.xml` runs end-to-end with output verified via h5py.

Part of #94.


abensonca (commented):
Reviewed - all looks good.

PR #1214, opened 2026-07-02
refactor: collapse remaining HDF5 type-duplicated functions using generics

### LINKED ISSUE #92: Use Fortran 2003 API for HDF5 read/writes

### LINKED ISSUE #94: Use generics to reduce duplicated code in the IO HDF5 module

abensonca:
## Purpose

Final step of the #94 deduplication campaign: sweep up the remaining type-duplicated pairs. Stacked on #1213 → #1212 → #1211 → #1210; merge in that order.

## What changed

**6 procedures → 3 templates** (net **−631 lines**), via two new generic directives:

- **`VlenType`** — the varying-length double and varying-length long integer 1-D dataset write/read pairs (`VarDouble_1D`/`VarInteger8_2D`) were verified to differ only in comments and trivial style (one nested-vs-combined `if`, `IO_HDF5_Open_Dataset(self,…)` vs `self%openDataset(…)`); they collapse into single write and read functions templated over {`hdf5VarDouble`, `hdf5VarInteger8`}. The VLEN marshaling through `hdf5VlenC` + `c_f_pointer` is fully element-type-agnostic.
- **`TableType`** — the real and integer table column readers were byte-identical apart from documentation (including both using `H5T_NATIVE_REAL` for the size lookup — now documented as intentional: only the 4-byte record size is used). They collapse into one templated reader.

Deliberately left hand-written: the long integer table reader (it uses the one remaining `H5TBread_fields_name` C-API wrapper, which we verified must survive at HDF5 1.14.5), the character variants, and the structurally distinct `VarVarDouble`/`VarDouble_2D` VLEN variants.

Also fixed en route: the `<generic>` directive schema requires `label` + `intrinsic` on every instance (caught by the pre-commit XML validation), so the new directives conform.

## Campaign totals

| | Start | Now |
|---|---|---|
| Module size | 18,340 lines | **8,473 lines (−54%)** |
| Hand-written read/write variants | ~90 | 72 collapsed into **11 templates** + specials |

Call sites were untouched throughout — every PR preserved the `writeAttribute`/`readAttribute`/`writeDataset`/`readDataset`/`readTable` generic bindings exactly.

## Verification

- All **210 assertions** pass (both VLEN round-trips, all four table column types, subsection table reads).
- Full `Galacticus.exe` build succeeds; `parameters/quickTest.xml` runs end-to-end.

Closes #94. Closes #92.

### DIFF
--- a/source/tests/IO/HDF5.F90
+++ b/source/tests/IO/HDF5.F90
@@ -32,48 +32,48 @@ program Tests_IO_HDF5
   use :: Units_MetaData    , only : unitType
   use :: Unit_Tests        , only : Assert             , Unit_Tests_Begin_Group, Unit_Tests_End_Group, Unit_Tests_Finish
   implicit none
-  integer                                                                   :: iPass                          , integerValue         , &
-       &                                                                       integerValueReread             , i                    , &
-       &                                                                       j                              , k
+  integer                                                                   :: iPass                            , integerValue                  , &
+       &                                                                       integerValueReread               , i                             , &
+       &                                                                       j                                , k
   logical                                                                   :: appendableOK
-  integer                                                                   :: unitsStatus                    , attributesStatus
+  integer                                                                   :: unitsStatus                      , attributesStatus
   type            (unitType       )                                         :: unitsValue
   integer                                       , dimension(10)             :: integerValueArray
   integer                                       , dimension(10)             :: integerValueArrayRereadStatic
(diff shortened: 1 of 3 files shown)


abensonca (commented):
Reviewed - all looks good.

Issue #1216, opened 2026-07-02
Double-free in varying_string finalization in the presetNamedReals property extractor


abensonca:
Running a model that outputs `presetNamedReals` node properties aborts with a glibc double-free during output:

```
./Galacticus.exe testSuite/parameters/bolshoiTestTreesGLC.xml
```

(the `bolshoiTestTreesGLC_in.hdf5` input having first been generated via `testSuite/parameters/mergerTreeFileBuildBolshoi.xml`).

Backtrace (gcc-16 build):

```
#10 malloc_printerr                                          malloc.c:5341
#11 tcache_double_free_verify                                malloc.c:3159
#12 __iso_varying_string_MOD_destructor_vs                   iso_varying_string.p.F90:2705
#13 __iso_varying_string_MOD___final_iso_varying_string_Varying_string
#14 __node_property_extractors_MOD_presetnamedrealsnames     preset_named_reals.p.F90:123
#15 __node_property_extractors_MOD_multinames                multi.p.F90:490
#16 __merger_tree_outputters_MOD_standardpropertynamesestablish
#17 __merger_tree_outputters_MOD_standardoutputtree
```

The crash is in the `varying_string` finalizer invoked from `presetNamedRealsNames` (`source/nodes/property_extractor/preset_named_reals.F90`), whose body is:

```fortran
    allocate(names(size(self%presetNames)))
    names='preset:'//self%presetNames
```

The elemental concatenation produces an array temporary of `varying_string` whose finalization appears to interact badly with the defined assignment, freeing the same buffer twice. A likely workaround is an explicit element-wise loop (`names(i)='preset:'//self%presetNames(i)`), which avoids the array temporary; other extractors with the same idiom may be similarly affected.

Notes:

* Reproduces identically on builds of current master-era code both before and after the recent merger-tree-read refactoring (discovered while using this model for a before/after comparison in the work tracked by #1215) — it is unrelated to that work.
* The crash is in the *output* path; removing the `presetNamedReals` entry from the `multi` `nodePropertyExtractor` allows the model to run to completion (the `presetNamedReals` option of `mergerTreeConstructor` "read" itself works fine).

abensonca:
Root cause identified: this is a **link-time-optimization miscompilation** (gfortran 16) of the whole-array elemental concatenation

```fortran
names='preset:'//self%presetNames
```

[...]

PR #1217, opened 2026-07-02
Simplify the merger tree read constructor

### LINKED ISSUE #1215: Simplification opportunities in the merger tree read constructor (merger_trees/construct/read/_class.F90)
*Code review of `source/merger_trees/construct/read/_class.F90`, 2026-07-01. Line references are to master at 82ac5cb3d. Stages 1–2 of the suggested order of work (low-risk cleanups and the split-forest removal) are implemented on the `feature/merger-tree-read-cleanup` branch; this issue tracks the remaining stages.*

*Review of the `mergerTreeConstructorRead` class (4,416 lines), 2026-07-01.*

## Overall

The file is doing the work of about four modules: core tree wiring (the part that truly
"must be correct"), property presetting (scale radii, spins, orbits, histories),
split-forest machinery, and diagnostics. The complexity that matters is concentrated in a
handful of repeated idioms — branch walks, progenitor surveys, clone insertion — each
hand-rolled at several sites with small variations. Consolidating those would shrink the
file substantially *and* remove the main correctness risk, which is drift between
near-copies that must stay mutually consistent.

## 1. Quarantine the split-forest machinery (~900 lines, currently unreachable)

`readConstruct` unconditionally calls `Error_Report` when a forest exceeds
`forestSizeMaximum` (`_class.F90:1065-1069`), and `testSuite/test-splitForests.py` is a
stub that prints "SKIPPED… currently broken". So `readRootNodeAffinitiesInitial`,
`readAssignSplitForestEvents`, the five push/pull-list methods, the split-forest half of
`readInterTreeMergeTimeSet`, and all the `splitForest*` state exist only to support a
dead path — roughly 20% of the file. Consider moving it to a sibling file (the type could
keep the bindings) or deleting it outright until the MPI work resumes; git preserves it
either way. Every reader of this file currently has to reason about push/pull lists
(e.g. the `isOnPushList` check inside `readBuildDescendantPointers:1518`) that can never
be populated.

## 2. Decompose `readConstruct` (465 lines)

Three extractions would leave a readable ~15-step pipeline:

- **Subhalo mass/angular-momentum accounting** (`1146-1224`): three separate node loops
  implementing a subtle divide-by-mass → add-subhalo-masses → multiply-back-by-mass trick
  for the `scale` method, plus the orbital-summation method. This invariant ("scalar spin
  is temporarily mass-normalized between these two loops") is exactly what deserves its
  own named method with a comment stating it.
- **Component-requirement validation** (`1237-1327`): ~90 lines of checks that depend
  only on `self%preset*` flags and the default components — constants for the whole run —
  yet they execute per tree. Hoist to the internal constructor alongside the other
  one-time checks (`772-901`).
- **The `beginAt` skip loop**: the `do while (.true.)` with a `cycle` that works by
  decrementing `treeNumberOffset` (`1021-1028`) is clever but opaque. A small
  `findNextForest()` helper returning the forest number or "finished" would make the
  control flow explicit.

## 3. Extract the four repeated idioms

**Clone-as-primary-progenitor insertion** appears four times:
`readScanForSubhaloPromotions:1699-1719`, `readValidateIsolatedHalos:3236-3260`, and
twice in `readAssignSplitForestEvents:4192-4217, 4230-4256`. All do copy → strip
satellite components → splice as `firstChild` → offset time by
`fractionOffsetTimeClones` → nullify `event` — but they differ in whether satellites are
transferred to the clone and how `sibling` is set. Some of those differences may be
intentional; extracting one `insertClonedProgenitor(node, transferSatellites=...)` would
force that question to be answered once and prevent future drift. Relatedly, clone
*detection* uses two idioms — `uniqueID()` equality (`2617`, `3451`, `3279`) and
`index()` equality (`2807-2810`) — which are only equivalent before
`readAssignUniqueIDsToClones` runs. A single `isClonedPrimary(node)` helper would encode
that ordering dependency.

**The subhalo branch walk** ("follow descendants until branch tip / merger with isolated
/ subhalo-subhalo merger / promotion") is implemented three times with different
bookkeeping: `readScanForMergers:2476-2537`, `readBuildSubhaloMassHistories:3116-3154`,
and `readScanForBranchJumps:2900-3015`. These *must* terminate identically — the runtime
check at `3120-3124` ("history array length exceeded… this should not happen") exists
precisely because `historyCountMaximum` from one walk must bound another walk's count. An
iterator in the style of the existing `progenitorIterator`, yielding each visited node
plus a termination reason, would make the consistency structural instead of asserted. It
would also allow dissolving the interleaved `passAssign`/`passMerge` two-pass structure
in `readScanForMergers`, which is currently the hardest-to-follow logic in the file.

**Progenitor surveys** ("does an isolated progenitor exist? is this node the most
massive?") are hand-rolled at six sites, and the tie-breaking is inconsistent.
`massIsGreater` (`2757`) exists specifically to break equal-mass ties by index so the
primary progenitor is well-defined — and `readIsSubhaloSubhaloMerger` and the walk bodies
use it — but `readScanForSubhaloPromotions:1675-1684` uses a `>`/`>=`-by-isolated-status
scheme, and `readScanForMergers:2451, 2499` and `readRootNodeAffinitiesInitial:3945-3949`
use plain `>`. For N-body trees with duplicated discrete masses these can disagree about
who is "most massive," meaning the promotion decision and the orbit-assignment decision
could pick different nodes. A single `progenitorSurvey(descendant)` returning
`(exist, isolatedExists, mostMassiveIndex)` with one documented tie-break rule would
replace all six loops. (Two of them, in the split-forest code, also reimplement iteration
on the raw sorted-index arrays instead of using `progenitorIterator`.)

**Orbit setting at the virial radius** — the propagate-or-random-fallback-or-assert block
is duplicated between `readSetOrbit:2722-2752` and `readInterTreeMergeTimeSet:4375-4405`.

Smaller but same category: the "walk to isolated host" loop
(`do while (associated(x%host) .and. .not.associated(x%host,x))`) appears ~8 times in
`readScanForBranchJumps`, `readDumpTree`, and the split-forest code; a one-line
`isolatedHost(node)` function would help.

## 4. Fragile spots worth hardening (possible latent bugs)

- **`readNodeLocation` returns 1 on lookup failure** (`1466-1468`). Callers must remember
  to verify `nodes(loc)%nodeIndex` matches; `readBuildDescendantPointers` does, but
  `readAssignMergers:2796` does not — a failed lookup there would silently treat
  `nodes(1)` as the merge target. Returning 0 (or erroring centrally with an `optional`
  found flag) would make misuse loud.
- **`progenitorIteratorNext:3680`** can evaluate `nodes(self%progenitorLocation)` with
  `progenitorLocation = -1` when the inner loop exhausts the sorted index
  (`progenitorIndex` reaches 0 at `3676` before a location is assigned). It's an
  out-of-bounds read whose result is discarded on the correct path, but it will trip
  `-fcheck=bounds` and is undefined behavior. Reordering the guards fixes it.
- **`isolatedNodeIndex` triple duty**: it holds the sentinel
  `nodeReachabilityUnreachable%ID`, the marker `nodeReachabilityReachable%ID` (`2389`),
  *and* a genuine 1-based index into `nodeList`. If the enum IDs are 0/1, "reachable" is
  indistinguishable from "belongs to isolated node 1"; correctness relies on
  `readScanForMergers` pass 1 overwriting every marker before anything reads it as an
  index. A separate logical field (or negative sentinels) would remove a whole class of
  ordering hazards, cheaply.
- **`readConstruct:1398-1402`**: `if (allocated(historyMass)) deallocate(historyIndex)` —
  the guard checks the wrong variable (copy-paste). Harmless today because all three are
  always allocated — which also means the entire block, and the pre-emptive
  `deallocate(position/velocity)` at `1373-1374` (never allocated at that point), can
  simply be deleted: they're locals.
- Minor: `readScanForMergers:2637` re-tests `alwaysPromoteMostMassive` inside an
  `if (self%alwaysPromoteMostMassive)`; `2630` initializes `iIsolatedNode=0` pointlessly;
  `forall` at `1440` is obsolescent (plain loop or `do concurrent`).

## 5. Dead code

`readTimingRecord`/`readTimingReport` (`3731-3783`) and the `timingTimes`/`timingLabels`
state are never invoked anywhere — either wire `timingRecord` calls into the pipeline
phases (they'd be genuinely useful for large forests) or drop them. `readDumpTree` is
deliberately kept as a debug tool (fine, though it could live in a debug utilities file).

## 6. The highest-leverage improvement: make the invariants checkable

The class documentation (`119-148`) is effectively a specification of pathological
topologies — initial satellites, isolated nodes with only subhalo progenitors, missing
hosts, branch jumps, promotions, forests. Two things would convert "must be correct" from
vigilance into verification:

1. **A post-construction invariant checker** (enabled under a debug flag or in the test
   suite): no cycles via descendant/host/parent; every subhalo's host chain terminates at
   a self-hosting node; times positive and non-decreasing along descendant links; every
   `mergeTarget` reachable and its `firstMergee` chain consistent; every node's
   `hostTree` set; clones have distinct `uniqueID`s. Most phases already contain
   scattered ad-hoc checks (`readBuildParentPointers`' self-parent check,
   `readEnforceSubhaloStatus`' failure scan) — this would centralize and extend them.
2. **Synthetic-tree unit tests**: tiny hand-built trees, one per documented topology, run
   through `construct` and checked against expected structure. Right now the pathological
   cases are exercised only implicitly by whatever real trees the test suite reads; a
   promotion-with-tie-mass case or a skipped-snapshot branch jump case would take ~20
   nodes each to encode and would pin down exactly the logic that's hardest to reason
   about.

## Suggested order of work

1. Low risk, mechanical: dead timing code, deallocate fixes, minor cleanups, iterator
   guard reorder, isolated-host helper, clone-insertion helper.
2. Split-forest quarantine (move to sibling file or remove).
3. `readConstruct` decomposition (hoist validation, extract angular-momentum accounting).
4. Progenitor-survey unification (fix tie-breaking inconsistency — behavior change,
   needs test-suite comparison).
5. Branch-walk iterator (dissolve the two-pass `readScanForMergers`).
6. Invariant checker + synthetic-tree tests.

abensonca:
Simplifies the merger tree read constructor (`source/merger_trees/construct/read/_class.F90`), implementing all six stages of the plan in #1215. The file shrinks from 4,416 to ~3,800 lines, with the complicated logic consolidated behind named, documented abstractions, and its correctness invariants now machine-checked.

The six commits, in order:

1. **Low-risk cleanups** — remove dead timing infrastructure; fix a wrong-variable deallocate guard; guard an out-of-bounds read in `progenitorIteratorNext`; extract `readIsolatedHost` (replacing 8 hand-rolled host walks) and `readInsertClonedProgenitor` (replacing 4 clone-splice blocks, with satellite handling made an explicit argument).
2. **Remove split-forest processing** (~700 lines) — the path has been unconditionally fatal pending MPI support and its test is a skip stub. `[forestSizeMaximum]` is retained; over-sized forests still produce a (now clearer) fatal error. Recoverable from git history.
3. **Decompose `readConstruct`** (~420 → ~220 lines) — extract `findNextForest` (the `[beginAt]` scan loop), `addSubhaloContributions` (with the normalize/add/denormalize mass-angular-momentum ordering invariant documented), and `validatePresetComponents`.
4. **Unify progenitor surveys** (⚠️ behavior change on ties) — six inconsistent hand-rolled surveys replaced by one `progenitorSurvey` method with a strict seniority ordering: descending mass, isolated halos outrank subhalos of equal mass, remaining ties broken by descending node index. Previously, equal-mass ties (common in N-body trees) were resolved inconsistently between sites and, in one case, iteration-order-dependently. Validated by before/after output comparison on six models: five bitwise identical (including real Bolshoi trees); the sixth changes only in trees from forests containing decision-relevant equal-mass ties.
5. **Branch-walk iterator** — the two copies of the subhalo branch walk (in `readScanForMergers` and `readBuildSubhaloMassHistories`, whose termination criteria must agree for history-array sizing) now share a `branchIterator`; the interleaved two-pass structure of `readScanForMergers` becomes two explicit loops. Verified behavior-preserving by the same output-comparison method.
6. **`validateStructure` merger tree operator + synthetic-tree tests** — a reusable operator checking structural invariants of constructed trees (linkage in both directions, bounded parent chains, merge target/mergee consistency, clone unique IDs, event pairing), plus `testSuite/test-merger-tree-read-ties.py` with two hand-built trees pinning the tie-resolution semantics of stage 4.

**Verification** (each stage): full rebuild; all 16 tree-reading regression models (including `lightconeCrossingTests.xml`); `test-mostMassiveProgenitorIsSubhalo.py`, `test-pruneLightcone.py`, `test-lightcone-crossing.py`; and, for stages 4–5, dataset-level before/after output comparison across six models.

Incidental findings, tracked separately: #1216 (double-free in the `presetNamedReals` property extractor, pre-existing); run-to-run nondeterminism of star formation history *mass* outputs under OpenMP (times and all scalar outputs are deterministic; not yet filed).

Closes #1215.


abensonca:
Pushed a fix (56c982328) for the CI failures (segfaults in `is_progenitor_of_node` and "multiple satellite components defined on branch" fatal errors across the `test-methods` read-constructor shards, `Model-Integration`, and `lightconeCrossingTests`).

**Root cause.** The progenitor-survey unification made the merger/promotion decisions use `readProgenitorIsSenior` (mass desc → isolated outranks subhalo on ties → node-index desc), but `readBuildChildAndSiblingLinks` still selected the primary progenitor (`firstChild` / `isPrimaryProgenitor()`) with a plain `mass > mass` comparison (first-encountered wins on ties). On trees with equal-mass progenitors the two disagree about which progenitor is primary, so a node could be simultaneously the primary progenitor of its parent and, per the survey, a merging node — leaving two satellite components on one branch. This only manifested during *evolution*, which is why construction and the earlier verification passed. It escaped my per-stage verification because I hadn't run `test-methods.py` locally.

**Fix.** Break mass ties in `readBuildChildAndSiblingLinks` by descending node index, matching `readProgenitorIsSenior`, so the primary-progenitor structure is consistent with the survey-based decisions. It only affects exact equal-mass ties (a no-op on tie-free trees).

**Verification** (clean build): the previously-failing `galacticus_8:17` config now runs to completion; the `validateStructure` operator reports no structural violations on the constructed `test-MergerTrees.hdf5` trees; and the full `test-methods.py` grid, all 16 tree-reading regression models, the tie-resolution tests, and the lightcone/promotion tests all pass. I've also added `test-methods.py` to my local verification recipe.

abensonca (commented):
Reviewed - all looks good (other than minor formatting fixes).

PR #1219, opened 2026-07-03
fix(build): propagate compiler, linker, and profiler exit statuses

abensonca:
**PR 1 of 3** (stacked: this → `fix/build-dependency-graph` → `clean/build-dead-code`), implementing the first item of the build-infrastructure review action plan.

## Problem

Three paths in the build masked hard failures, letting `make` record failed steps as successful:

1. **Compile recipes** piped compiler output into `postprocess.py`, so the recipe's exit status was the postprocessor's (fails only on `^Error:` lines), not the compiler's. An ICE, segfault, or OOM kill emits no such line → the recipe "succeeded" and the missing object surfaced later as a confusing link error.
2. **Link recipes** piped through `postprocessLinker.py` the same way — and real link failures print `undefined reference` / `collect2: error: ld returned 1 exit status`, neither matching `^Error:`, so a failed link exited 0 and make recorded the executable as built.
3. **`profiler.sh`** (the SHELL under `GALACTICUS_BUILD_OPTION=compileprof`) discarded the wrapped command's status entirely and always exited 0 — make could not detect *any* failed recipe during a profiled build (CI survives today only by grepping the build log afterwards).

## Fix

Compiler/linker diagnostics now go to a temporary file which is postprocessed after capturing the tool's own exit status; the recipe fails if either the tool or the postprocessor reports failure. Applied to the Makefile `%.o` rule, the mirrored wrapper-unit recipe emitted by `libraryInterfacesDependencies.py` (sync-obligation comments added on both sides), and the link recipes generated by `findExecutables.py`. `postprocessLinker.py` additionally flags `error:` (case-insensitive), `undefined reference`, and `ld returned N exit status` lines as a second line of defense. `profiler.sh` captures and propagates the wrapped command's status.

## Verification

- Failure injection at each level: missing input file, type error, ICE-style silent failure (`f951: Fatal Error` with no `Error:` line), synthetic linker errors, and a failing recipe under the profiler SHELL — all now fail the recipe; clean compiles/links still pass, and the filtered-warning cases still exit 0.
- Clean full rebuild (`make -j20 Galacticus.exe`, 1,796 objects) and an incremental rebuild both succeed.

Note: merging this invalidates all existing objects (the `%.o` recipe changed and every object depends on the Makefile), so everyone's next build is a full rebuild.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -305,9 +305,22 @@ FCFLAGS_AARCH64_ICE_OBJECTS = \
 $(FCFLAGS_AARCH64_ICE_OBJECTS): FCFLAGS += -O0
 endif
 
+# The compiler writes diagnostics to a file which is then fed to postprocess.py, rather than piping
+# compiler output directly into postprocess.py. In a pipeline the compiler's exit status would be
+# discarded (the recipe's status would be postprocess.py's, and /bin/sh has no pipefail), so a hard
+# compiler failure that emits no "Error:" line — an internal compiler error, a segfault, an OOM
+# kill — would be reported as success, leaving a missing or stale object file that only surfaces
+# later as a confusing link error. Capturing and testing both statuses makes such failures fail the
+# recipe immediately. (libraryInterfacesDependencies.py emits a mirror of this recipe for the
+# library wrapper units — keep the two in sync.)
 $(BUILDPATH)/%.o : $(BUILDPATH)/%.p.F90 $(BUILDPATH)/%.m $(BUILDPATH)/%.d $(BUILDPATH)/%.fl Makefile
 	@mkdir -p $(BUILDPATH)/moduleBuild
-	$(FCCOMPILER) -c $(BUILDPATH)/$*.p.F90 -o $(BUILDPATH)/$*.o $(FCFLAGS) 2>&1 | ./scripts/build/postprocess.py $(BUILDPATH)/$*.p.F90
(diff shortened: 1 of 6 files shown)


abensonca:
Pushed a follow-up commit fixing the macOS CI failures (Build-Executable-MacOS / -M1, Build-Tools-MacOS-M1).

**Root cause:** the macOS static relink runs `staticRelinker.py $(grep '-o <exe>' build.log)`, feeding the script the link command line extracted from the build log. This PR's recipe change moved linker diagnostics from a pipe (`… 2>&1 | ./scripts/build/postprocessLinker.py`) to a redirect (`… > <stem>.link.diag 2>&1; \`), and `staticRelinker.py` only knew how to split at `2>&1` and re-append the tail — so it treated `> <stem>.link.diag` as link arguments and re-appended a broken `2>&1 ; \`, producing a malformed shell command that exited 127. The relink itself was fine; only the recipe-line parsing broke. (Exactly the exit-code interaction anticipated in review.)

**Fix:** `staticRelinker.py` now stops parsing at the first shell operator (redirect / pipe / terminator / trailing line-continuation) and discards the remainder — it reconstructs and runs its own link command and never needed the recipe's diagnostics plumbing. Robust to both the old and new recipe shapes; verified by simulating the grepped line through the parser for both. (Also cleaned up two pre-existing invalid-escape warnings on the same relink code path.)

The other 8 PRs in the stack were rebased onto this and force-pushed; bases unchanged.

abensonca (commented):
Reviewed - all looks good.

PR #1220, opened 2026-07-03
fix(build): close dependency-graph gaps and repair build-cache tracking

abensonca:
**PR 2 of 3**, stacked on #1219 (`fix/build-status-propagation`); merge that first — GitHub will retarget this PR to `master` when its base branch is deleted. Implements items 2 and 3 of the build-infrastructure review action plan.

## Dependency-graph fixes

- **Directive catalogs**: `codeDirectivesParse.py` / `stateStorables.py` / `deepCopyActions.py` now run under a `directiveCatalogs.stamp` rule listing all three scripts as prerequisites (previously only the first was listed, so edits to the other two never regenerated their catalogs). Their outputs — `Makefile_Directives`, `directiveLocations.xml`, `stateStorables.xml`, `deepCopyActions.xml` — were side effects with no rules of their own despite being named as prerequisites elsewhere; each now has an empty-recipe rule on the stamp. Side benefit: the stamp stops the catalog scripts re-running on every invocation — a no-op `make deps` drops to ~2s / "Nothing to be done".
- **OpenMP critical-section enumeration**: previously depended only on its own script, so adding/removing/renaming a named `!$omp critical` never regenerated it until the build dir was wiped. Now depends on all Fortran sources via a `.up` sentinel; the script uses the shared `FileChanges.update` (replacing its private copy), touching the sentinel every run while rewriting the `.xml` only-if-changed so no-op regenerations don't cascade.
- **Preprocessing** now depends on `stateStorables.xml`/`deepCopyActions.xml`, which the preprocessor's hooks read at run time — previously a catalog change didn't invalidate already-preprocessed outputs (a classic "wipe the build tree" cause). Made safe by converting `deepCopyActions.py` to atomic only-if-changed writes (matching `stateStorables.py`); it also now skips files deleted since `directiveLocations.xml` was generated instead of crashing.
- `python/LibraryInterfaces/Hierarchy.py` added to the `libgalacticus.Inc` prerequisites (it's imported by `libraryInterfaces.py`); `Makefile_Directives` is written only-if-changed so make stops re-executing itself after no-op regenerations.

## Cache-tracking fixes

- **`sourceDigests.py`**: `_append_dep_files` still assumed the pre-hierarchy flat source layout — it kept only basenames, so its `.d`-sidecar and source lookups failed for any file in a subdirectory (essentially all files) and it silently tracked no dependencies. A type's `sourceMD5` was therefore not recomputed when a dependency changed, quietly defeating the state-restore safety check the digests exist for. Paths now preserve subdirectories; the blob cache is versioned (old caches discarded at a one-time full-rescan cost per target); entries record the types they provide; stale entries/types are pruned, with pruned types feeding composite invalidation.
- **`parameterDependencies.py`**: never pruned cache entries, so parameters from files that left an executable's `.d` kept inflating `knownParameterNames` forever, weakening unknown-parameter detection. Entries are now pruned each run, and the `.p` parameter-list sidecar is tracked for staleness.

## Verification

- New regression tests: `scripts/build/tests/test_source_digests.py` (hierarchical-path handling; would have caught the flat-layout bug).
- Clean full rebuild (`make -j20 Galacticus.exe`); incremental rebuild after touching a core module re-preprocesses exactly that one file, all catalog mtimes stay stable, no re-preprocess cascade.

Note: the digest-blob version bump forces a one-time digest recomputation per target on the first build after merging.


abensonca (commented):
Reviewed - looks good.

PR #1221, opened 2026-07-03
clean(build): remove dead code from the build infrastructure

abensonca:
**PR 3 of 3**, stacked on #1220 (`fix/build-dependency-graph`); merge that first — GitHub will retarget this PR when its base branch is deleted. Implements item 4 of the build-infrastructure review action plan: pure dead-code removal plus a few clarifying comments (net −39 lines; no behavior change on any live path).

Each removal was verified unreachable or unreferenced before deletion:

- **Makefile**: the generic `%.exe` pattern rule — every executable gets an explicit rule from `findExecutables.py`, and the pattern rule's backtick/`$*` prerequisites can't work without `.SECONDEXPANSION`, so it could never fire (a comment now points readers at the generated rules). With it go `CONDORLINKER` (Condor standard universe is long discontinued; that rule was its only user) and the `aux/XSpec` rule (`aux/XSpec/` doesn't exist). `FCFLAGS_NOOPT` is kept but documented as provenance-only.
- **findExecutables.py**: `$(UPDATE_DEPS)` (defined nowhere) dropped from generated prerequisites; unreachable `t?` in the stem regex.
- **moduleDependencies.py**: inert include-following — its `include\s+'(\w+)'` regex could not match any real include in the tree (all include names contain a `.`, which `\w` excludes), so the branch never fired; plus an unused parameter and stale docstring text.
- **useDependencies.py**: the near-unreachable, broken-if-reached `subprocess.run(['make', leaf])` missing-file fallback; a pointless alias.
- **buildCode.py**: the fallback tree walk (the Makefile ordering guarantees `directiveLocations.xml` exists; a missing map is now a clear out-of-sequence error), a dead `except StopIteration` and the unreachable return behind it, and a stale header comment still citing `buildCode.pl`.
- **buildProfiler.py**: Perl-era command simplifiers (no Perl remains in the build).
- **includeDependencies.py**: unused regex group; meaningless `automatic` flag on header entries (the flag's real meaning — the `! NO_USES` opt-out — is now documented where it's set).
- **libraryInterfaces.py**: unused imports, stale `sys.path`-shim comment, dead `required=False` branches in `_load_xml`.
- **postprocess.py**: the pointless `try: import sys as _sys` construct.
- **codeDirectivesParse.py**: redundant blob-existence re-check; the currently-unused `<base>.function`/`.moduleUse`/`.functionCall` branches are kept but documented as future-proofing.
- **preprocess.py**: `GALACTICUS_PREPROCESSOR_ANALYZE` documented as a manual-only debugging hook.
- **libraryInterfacesAudit.py**: unused import; docstring now states it is a manual developer tool invoked by nothing in the Makefile or CI, and warns that its deliberately duplicated classification predicates can silently drift from the generator (they already have, on `omp_lock_kind` — de-duplication is a planned follow-up; deleting the script outright remains an option).

## Verification

Clean full rebuild (`make -j20 Galacticus.exe`, 1,796 objects) and incremental rebuild with the complete three-PR stack applied; the stacked tip is byte-identical to the tree those builds ran on.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -92,10 +92,6 @@ else
 CPPCOMPILER ?= g++
 endif
 
-# Linker for Condor standard universe executables. Uncomment the second line to link for submission to the Condor standard universe.
-CONDORLINKER = 
-#CONDORLINKER = condor_compile
-
 # Fortran compiler flags:
 FCFLAGS += -ffree-line-length-none -frecursive -DBUILDPATH=\'$(BUILDPATH)\' -J$(BUILDPATH)/moduleBuild/ -I$(BUILDPATH)/ ${GALACTICUS_FCFLAGS} -pthread
 # Fortran77 compiler flags:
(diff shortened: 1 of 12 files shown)


abensonca (commented):
Reviewed - looks good.

PR #1222, opened 2026-07-03
refactor(build): give the link recipe and configuration probes single owners

abensonca:
**PR 4 of 7** in the build-infrastructure series, stacked on #1221 (`clean/build-dead-code`); merge in order — GitHub retargets each PR when its base branch is deleted. Implements item 5 of the review action plan, plus two latent-bug fixes its verification exposed.

## Deduplication

- The full executable-link procedure (parameter enumeration → MD5 source digests → status-checked link) previously existed in three drifting copies. It now lives in two canned recipes in the Makefile — `LINK_METADATA` (shared by executables and `libgalacticus.so`) and `LINK_EXECUTABLE` — and `findExecutables.py` emits just a dependency line plus a one-line `$(call ...)` per executable (`Makefile_All_Execs` drops from ~22 lines per executable to 2). Changing the link procedure now means editing one place.
- The five identically-shaped availability probes (OFD, FFTW3, ANN, QHull, MathEval) share a `CONFIG_PROBE` canned recipe; adding a probe is a `-include` line, a rule, and one `$(call)`. Compiler and flags are passed by name (computed variable references) since flag values may contain commas. The differently-shaped probes (Proc, Git2) remain hand-written.

## Bug fixes exposed during verification

- **`qhull.o` staleness**: the hand-written qhull rule was the only object rule without `Makefile` in its prerequisites. `qhull.cpp` preprocesses to an empty translation unit under `-DQHULLUNAVAIL`, so when the availability result flips, every other object rebuilds while the stale stub survives — and the link fails with `undefined reference to convexHullVolumeC`. Exactly this happened during stack verification.
- **Probe poisoning** (the `git2.h` trap, striking again): the ANN/qhull/matheval probes compiled with the full `$(CPPFLAGS)`, whose `-I$(BUILDPATH)/` can contain zero-byte header stubs left by an earlier UNAVAIL build (created by the generic `%.h` rule). A stub satisfies the probe's `#include`, flipping it spuriously to AVAIL on a machine without the real headers — after which the real compile fails on the empty header. The three probes now compile against `GALACTICUS_CPPFLAGS` only, mirroring the libgit2 probe that hit the identical trap first.

## Verification

Expanded recipes verified semantically identical to the previous generated rules (dry-run comparison); all probes re-verified (including that the poisoned QHULLAVAIL reverts to an honest UNAVAIL once the stub is out of the probe path); `tests.sort.exe` built end-to-end through the canned recipes; full `make -j20 Galacticus.exe` at the stack tip (1,796 objects) links cleanly. Only output change: `Makefile_Config_*` fragments lose column-alignment padding.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -398,20 +398,29 @@ $(BUILDPATH)/Makefile_Config_OFD: source/system/flock_config.c
 $(BUILDPATH)/Makefile_Config_FFTW3: source/external/FFTW/fftw3_config.F90
 	$(call CONFIG_PROBE,FFTW3,FCCOMPILER,source/external/FFTW/fftw3_config.F90,FCFLAGS,FCFLAGS,FFTW3)
 
+# The ANN, qhull, and libmatheval probes compile against only the system/user headers
+# (GALACTICUS_CPPFLAGS), *not* the full CPPFLAGS. The latter adds `-I$(BUILDPATH)/`, which can
+# contain zero-byte header stubs left by an earlier UNAVAIL build (created by the generic `%.h`
+# rule for headers that were preprocessed out — e.g. `Qhull.h`). Such a stub satisfies the probe's
+# `#include`, making it spuriously report AVAIL even though the real headers are absent, and the
+# subsequent real compile then fails (or the availability flip-flops between builds). This mirrors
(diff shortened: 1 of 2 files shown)


abensonca (commented):
Reviewed - looks good.

PR #1223, opened 2026-07-03
refactor(build): extract the build scripts' shared helpers into single owners

abensonca:
**PR 5 of 7**, stacked on #1222. Implements item 6 of the review action plan in three commits, each independently verified as behavior-neutral (except where noted — the audit tool was repaired from a fully broken state).

## Commit 1 — shared cache/type-scan/hook-registration helpers

- `Galacticus.Build.ScanCache` (`file_identifier` + `load_cache`): previously cloned in six scripts and inlined in two more; its docstring now documents the blob freshness protocol as a whole.
- `Galacticus.Build.TypeScan`: the derived-type scanner shared by `deepCopyActions.py`/`stateStorables.py` (whose copies cross-referenced each other).
- `Galacticus.Build.SourceTree.Process.all`: single registration point for every process hook — `preprocess.py` and `buildCode.py` carried identical 32-line import blocks where a hook added to one but not the other would silently generate different code.

## Commit 2 — one library-knowledge registry

`useDependencies.py` knew which modules/includes imply a library requirement; `libraryDependencies.py` knew the dependency graph, static-link ordering, and availability conditions — with no cross-reference. All six tables now live in `Galacticus.Build.Libraries` with a docstring explaining which script consumes which half.

## Commit 3 — shared generator/audit classification (and audit repair)

The audit's hand-synced copies of the generator's classification rules (marked "MUST stay in sync") had drifted — the generator's `omp_lock_kind` rejection had no audit counterpart. The rules now live once in `LibraryInterfaces.Classification`: the generator imports them; the audit's `classify_constructor` delegates to the shared predicate in an audit mode that additionally distinguishes missing dependencies from pipeline blockers. Also repairs two rot bugs that had left the (never-exercised) audit completely broken: a flat `glob('*.F90')` from before the source-tree hierarchization (it discovered **zero** functionClasses and crashed on a division by zero), and a directive regex predating the `docformat` attribute.

## Verification

Byte-identical generated outputs at every step: all dependency makefiles and catalog XMLs; the emitted link flags for two executables; and a full `GALACTICUS_BUILD_OPTION=lib` generation run (libgalacticus.Inc, all 1,518 wrapper units, galacticus.py, and the skip-warning set). The repaired audit's classifications are unchanged by the delegation (report diff shows only generator-phrased blocker messages).


abensonca (commented):
Reviewed - looks good.

PR #1224, opened 2026-07-03
perf(build): eliminate redundant reads, parses, and rewrites in the build scripts

abensonca:
**PR 6 of 7**, stacked on #1223. Implements item 7 of the review action plan in three commits.

- **Library-interface generation**: each implementation file is parsed once (two walks of one tree) instead of twice — parsing dominates the generator's cost, and a full generation run roughly halves. The per-class/per-impl wrapper `.F90` units are written only-if-changed, so an unrelated catalog change no longer cascades into re-preprocessing and recompiling all ~1,500 wrapper units (`libgalacticus.Inc` itself stays an unconditional write — it is the target of the rule that runs the generator, so its mtime must advance). Prerequisite fix shipped first: the stale-wrapper sweep (and the unit enumeration in `libraryInterfacesDependencies.py`) matched live `<unit>.p.F90` files; both were masked by the unconditional rewrites and would have deleted live build products once write-if-changed landed.
- **useDependencies**: `extract_directives` now accepts a set of names, and the per-file worker makes one call instead of eight (each call is a full file read — nine reads per file per scan, counting the main scan). The functionGlobal module list and containing-module lookups are memoized per worker process instead of being re-derived (with full file re-reads) once per requesting source file. A full uncached rescan of the source tree drops to ~5s.
- **postprocess.py**: exits immediately when the compiler emitted no diagnostics (the common case, previously paying a full pre-scan of the preprocessed source per object file), and uses binary search over the line map.

One review sub-item deliberately deferred: parallelizing `sourceDigests.py` — its `find_hash` `.md5` sidecar writes would race *within* a run whenever `useLocks=no`, so it needs the locking design reworked first.

## Verification

Byte-identical outputs for every refactored path: full lib-generation run (all 1,518 wrappers), `Makefile_Use_Dependencies` from a full uncached rescan, and postprocess output/exit-status A/B on a failing compile. A planted `<unit>.p.F90` survives the fixed sweep; a no-op generator rerun rewrites zero files.

### DIFF
--- a/python/Galacticus/Build/Directives.py
+++ b/python/Galacticus/Build/Directives.py
@@ -41,12 +41,17 @@ def _matches_conditions(directive: dict, conditions: dict | None) -> bool:
     return True
 
 
-def extract_directives(file_name: str, directive_name: str,
+def extract_directives(file_name: str, directive_name,
                        conditions: dict | None = None,
                        set_root_element_type: bool = False) -> list[dict]:
     """Return every directive in `file_name` whose root element matches
-    `directive_name` (or whose root element is anything when `directive_name`
-    is `'*'`).
+    `directive_name` — a single name, a set/tuple/list of names, or `'*'`
+    to match any root element.
+
+    Passing several names in one call reads and parses the file once
+    (each call performs a full line-by-line read, so N single-name calls
+    cost N reads); use `set_root_element_type=True` to recover which name
+    each returned directive matched.
 
     Mirrors Perl Extract_Directives (Directives.pm:80-98) plus the underlying
     state machine from Extract_Directive (Directives.pm:11-78):
@@ -124,7 +129,7 @@ def extract_directive(file_name: str, directive_name: str, **kwargs: Any) -> dic
     return directives[0] if directives else None
 
 
-def _parse_xml_block(xml_text: str, file_name: str, directive_name: str,
+def _parse_xml_block(xml_text: str, file_name: str, directive_name,
                      conditions: dict | None,
                      set_root_element_type: bool) -> dict | None:
     """Parse one accumulated XML block and return a dict if the root matches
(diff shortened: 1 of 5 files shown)


abensonca:
The `Build-Executable-Linux-Non-Static` failure is fixed — root-caused to a regression introduced earlier in the stack (#1220), and the fix is committed there and rebased through this PR.

**Root cause:** #1220's `directiveCatalogs.stamp` restructuring made `Makefile_Directives` an `@true` target whose real regeneration lived on a stamp prerequisite, and had `codeDirectivesParse.py` write it only-if-changed. `Makefile_Directives` is `-include`d, and GNU make only re-executes itself (the restart that re-reads a regenerated included makefile) when that makefile is remade **by its own rule** running and updating it. With the work on the stamp + an `@true` recipe (and a stable mtime), make never restarted, so it never re-read the freshly generated rules for the generated `include`-directive files (the node-component include) nor the ordering line making `Makefile_Use_Dependencies` depend on them. `useDependencies.py` then ran once, before those includes existed, and silently dropped every `use` they contribute — surfacing only as the link-time `undefined reference` failures to `__node_component_*_data_MOD_*` / `object_pools` you saw, and only in a clean `make -j` build (a warm/incremental build regenerates identically, which is why it escaped the byte-identical verification).

**Fix (in #1220):** `Makefile_Directives` is a direct target with the catalog recipe again (keeping #1220's real fix that all three catalog scripts are prerequisites) and is written unconditionally. Verified by a clean `make -j4 LTO=disabled tests.tree_branch_destroy.exe` (this job's failing target): zero undefined references, no "No rule to make" errors, matching a master build. Bisected with a fast probe — the node-component dependency count for `objects/nodes/_class.o` was 5 on master/#1219, 0 at #1220, and is restored to 5.

abensonca (commented):
Reviewed - looks good.

PR #1225, opened 2026-07-03
fix(build): make useDependencies' conditional-compilation parsing sound

abensonca:
**PR 7 of 7**, stacked on #1224. Implements item 8 of the review action plan.

The two state machines deciding which conditionally compiled code the dependency scanner honors mis-parsed real inputs:

**Makefile-conditional scanner** — `ifneq` was unrecognized: it pushed nothing, but its `endif` still popped, desynchronizing the stack and randomly (de)activating enclosing branches; `else` was ignored entirely; `ifdef` names were uppercase-only and `-D` macro names uppercase-and-digits (a future `-DUSE_MPI` would truncate to `USE`). The stack now pushes on every opener and pops on every `endif`, with three-valued entries: known-active / known-inactive (ifdef/ifndef, evaluated against the environment) / unevaluable (`ifeq`/`ifneq`/chained `else if…`), whose branches are all collected — deliberate over-approximation, since over-depending is benign while under-depending silently drops dependencies.

**Source-side preprocessor stack** — `#if <expression>` blocks were treated as *inactive*, silently dropping any `use` statement inside them from the dependency graph; they are now unevaluable-active (every branch scanned). `#elif` is recognized (degrading its chain to unevaluable from that point, keeping an evaluable leading `#ifdef` exact), and `#ifndef` accepts lowercase macro names (`#ifndef __aarch64__` previously left its `#endif` to unbalance the stack).

The `Makefile_Use_Dependencies` blob is version-stamped so caches built under the old rules are discarded (one ~5s full rescan on first build after merging).

## Verification

Adds `scripts/build/tests/test_use_dependencies_conditionals.py` (10 tests) pinning both machines, including regression cases for the `ifneq` desync, the dropped `#if` dependencies, and lowercase `#ifndef`. On the current source tree the regenerated dependency graph is byte-identical (today's `#if` blocks carry no unique `use` statements) — i.e. no regression — and the full `make -j20 Galacticus.exe` at this stack tip builds and links cleanly with all 174 build-script tests passing.

### DIFF
--- a/scripts/build/tests/test_use_dependencies_conditionals.py
+++ b/scripts/build/tests/test_use_dependencies_conditionals.py
@@ -1,0 +1,231 @@
+"""Tests for useDependencies.py's conditional-compilation handling.
+
+Covers the two state machines that previously mis-parsed real inputs:
+
+* the Makefile-conditional scanner (`_collect_preprocessor_directives`),
+  which desynchronized its stack on `ifneq` (no push, but its `endif`
+  still popped) and ignored `else`;
+* the source-side preprocessor stack, which treated `#if <expression>`
+  blocks as *inactive* (silently dropping the `use` statements inside
+  them), did not recognize `#elif`, and rejected lowercase `#ifndef`
+  macro names.
+"""
+
+import os
+import sys
+
+import pytest
+
+sys.path.insert(0, os.path.join(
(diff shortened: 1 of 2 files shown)


abensonca (commented):
Reviewed - looks good.

PR #1226, opened 2026-07-03
docs: document the full build system in the developer guide

abensonca:
**PR 8 of 8** in the build-infrastructure series, stacked on #1225 (it documents the system as it stands at the top of the stack, so it must merge last). Implements item 9 of the review action plan.

The developer guide's build-system chapter covered the dependency-discovery layer but omitted roughly half the system. This extends `development.rst` (224 → ~490 lines) to cover, with all existing section labels preserved:

- **A four-phase build overview** (catalog → preprocess → compile → link) so a new maintainer has a map before the detail.
- **Every configuration knob** — `LOCKMD5`, `LTO`, `USEGIT2`, `SUFFIX`, `GALACTICUS_OBJECTS_DEBUG`, `OFDLOCKS`, compilers and flag variables — previously five of these had zero documentation anywhere. Plus the availability probes (including the header-stub poisoning caveat from #1222 and the "probes cache until their source changes — delete `Makefile_Config_*` after installing a library" gotcha) and the canned link recipes.
- **The incremental-build machinery as a coherent design**: the `.up` sentinel protocol, the blob-cache freshness rules, the `.mod` compare-and-swap, and the directive-catalog stamp — each existed only as fragments in code comments.
- **A source-marker reference**: every magic comment the build scripts key on (`!![`/`!!{` blocks, `!/ exclude`, `! NO_USES`, `!;`, `!:`/`//:`, `!$GLC` variants, `!-->`) and the hard-wired dependency implications of `!$omp critical`/`!$omp parallel`/`program` — none were written down before.
- **New discovery sections**: state storables & deep-copy actions, OpenMP critical sections, source digests (with the `LOCKMD5` locking story), on-the-fly functionClass submodules, and the use-scan's conditional-compilation semantics (from #1225).
- **Code generation updates**: exit-status propagation through the diagnostics files (from #1219), the `.lmap` format spec, and `Process.all` hook registration (from #1223).
- **The shared-library interface pipeline** and the parameter catalog/schema targets.
- **An extended Build Files inventory** — also correcting a stale claim (parameterDependencies writes `.parameters.F90`, not `.parameters.xml`).
- **The `++Task:` format contract** between `profiler.sh` and `buildProfiler.py`, and a historical note making the ported scripts' `<name>.pl:<line>` docstring references resolvable from git history.

## Verification

Full local sphinx build: no new warnings (the `GraphViz` `:term:` warnings pre-exist; `contributors.rst` is CI-generated), rendered heading hierarchy checked, and all cross-reference labels resolve through the `galacticus-ref` role (bare-label anchors verified against its resolver in `conf.py`).

### DIFF
--- a/docs/manuals/developer-guide/development.rst
+++ b/docs/manuals/developer-guide/development.rst
@@ -6,35 +6,209 @@ Developing Galacticus
 The Galacticus Build System
 ---------------------------
 
-Galacticus use a GNU Make-based build system. Dependencies are automatically discovered, and extensive meta-programming and preprocessing of source files is undertaken as part of the build. As a result, the build system is complicated. In this section the build system is described and detailed.
+Galacticus uses a GNU Make-based build system. Dependencies are automatically discovered, and extensive meta-programming and preprocessing of source files is undertaken as part of the build. As a result, the build system is complicated. This section describes it in detail: the Makefile and its configuration knobs, the incremental-build machinery, the special comment markers that the build scripts read from source files, the automatic-discovery and code-generation steps, and the files everything produces.
+
+A build proceeds, conceptually, in four phases:
+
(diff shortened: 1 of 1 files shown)


abensonca (commented):
Reviewed - looks good.

PR #1227, opened 2026-07-03
test(build): pin the build system's load-bearing invariants

abensonca:
**PR 9 of 9** — the final PR of the build-infrastructure series, stacked on #1226. Implements item 10 (the last item) of the review action plan: tests for the shared plumbing whose invariants were previously enforced by nothing.

57 new tests, in the review's priority order:

- **`ParallelScan`** — the byte-identical-to-serial and task-order contracts that keep every generated makefile deterministic under any parallelism; the `GALACTICUS_BUILD_JOBS` → `MAKEFLAGS -j` → standalone-default worker-count precedence (bare `-j`, no `-j` under make, clamping); the fail-with-serial-repro-hint behavior.
- **`FileChanges.update`** — only-if-changed replacement with mtime preservation, and the `.up` sentinel advancing on every run even when the output is untouched: the backbone of incremental-build minimality documented in #1226.
- **`ScanCache`** — stable cache keys and graceful degradation on missing/corrupt/wrong-type blobs.
- **`_canonicalize` pickle-identity** (codeDirectivesParse) — the most over-clever construct in the build scripts, whose "keep `_ENTRY_LITERALS` in sync with `_scan_one`" obligation had no enforcement. The test drives the real `_scan_one` over fixtures exercising every top-level entry shape (functionClass directive with implicit task dependencies, plus an include directive), simulates the worker pickle round-trip, and requires the canonicalized blob to pickle byte-identically to a serial one — with a guard asserting the round-trip genuinely de-interns, so the test cannot silently go vacuous.
- **`LibraryInterfaces.Classification`** — table-driven coverage of the shared argument rules from #1223, including the `omp_lock_kind` rejection (the rule whose silent loss from the audit's hand-synced copy motivated sharing the module), generator-mode vs audit-mode `class(...)` handling, overrides, hierarchy resolution, and the array-shape matrix.
- **The `++Task:` profiler contract** — end-to-end subprocess tests that profiler.sh's output parses with buildProfiler.py's regex and renders to a report, plus its exit-status propagation from #1219 (skipped where GNU `date` is unavailable, e.g. macOS).

The profiler test immediately earned its keep: `buildProfiler.py` divided by zero on any log whose tasks all completed within one second (timestamps have 1s resolution) — fixed here.

Full suite: 356 tests passing.

This completes the review action plan: with this PR merged, all ten items are done.

### DIFF
--- a/python/Galacticus/Build/tests/test_file_changes.py
+++ b/python/Galacticus/Build/tests/test_file_changes.py
@@ -1,0 +1,62 @@
+"""Tests for Galacticus.Build.FileChanges.update — the only-if-changed file
+replacement that (with its ``.up`` sentinels) is the backbone of the build's
+incremental minimality: a generated file's mtime must advance only when its
+content actually changes.
+"""
+
+import os
+
+from Galacticus.Build.FileChanges import update
+
+
+def _write(path, text):
+    path.write_text(text)
+    return str(path)
+
+
+def test_missing_old_moves_new_into_place(tmp_path):
+    new = _write(tmp_path / 'x.tmp', 'content')
+    old = tmp_path / 'x'
+    update(str(old), new)
+    assert old.read_text() == 'content'
+    assert not os.path.exists(new)
+
+
+def test_identical_preserves_mtime_and_removes_new(tmp_path):
+    old = _write(tmp_path / 'x', 'same')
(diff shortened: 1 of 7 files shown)


abensonca (commented):
Reviewed - looks good.

PR #1228, opened 2026-07-03
fix(build): read C/C++ sources one line at a time in buildCode

abensonca:
Stacked on #1227 (the tip of the build-infrastructure series). Fixes a latent correctness bug the review flagged but the plan deferred.

`buildCode.py`'s `_process_until_include_or_eof` read C/C++ files with `(fh.readline(), fh.readline(), '')` — which evaluates `readline()` **twice**, consuming two lines per iteration. So `raw_line` and `processed_line` were different, consecutive lines and every other line was skipped. A directive block spanning three consecutive lines (`!![` / `<tag .../>` / `!!]`) was missed entirely: the opener set `in_xml` on one line, `!!]` cleared it two lines later, and the `<tag>` line in between was never examined.

The C/C++ branch now reads one line and uses it for both the raw and processed forms — matching `_consume_until_close`, which already does this correctly a few lines below, and the Perl original. The Fortran branch (which relies on `get_fortran_line`'s continuation-joining) is untouched.

**Latency:** not currently triggered — the only include directive (`component`) lives in Fortran sources — but a real bug for any C/C++ source carrying a directive, which `codeDirectivesParse.py`'s directive-location map can include.

**Tests:** adds `scripts/build/tests/test_build_code_c_scan.py`, which drives the real scanner over C fixtures whose directive straddles the two-line read window. I verified these fixtures return no directives against the buggy code and the correct directive against the fix (a genuine regression guard, not a vacuous one). The Fortran path was re-verified end-to-end by regenerating the 5.3 MB component include, and the full build-script suite passes (234 tests).

### DIFF
--- a/scripts/build/buildCode.py
+++ b/scripts/build/buildCode.py
@@ -293,9 +293,18 @@ def _process_until_include_or_eof(fh, frame, build, source_directory,
     we exhausted the file.
     """
     while True:
-        raw_line, processed_line, _ = get_fortran_line(fh) \
-            if build['codeType'] == 'fortran' \
-            else (fh.readline(), fh.readline(), '')
+        if build['codeType'] == 'fortran':
+            raw_line, processed_line, _ = get_fortran_line(fh)
+        else:
+            # For C/C++ the raw and processed forms are the same single line;
+            # `get_fortran_line`'s continuation-joining does not apply. Read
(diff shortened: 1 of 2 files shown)


abensonca (commented):
Reviewed - looks good.

PR #1229, opened 2026-07-03
fix(build): keep all instances when one file implements two functionClasses

abensonca:
Stacked on #1228. Fixes the last of the review's deferred correctness bugs.

`stateStorables.py`'s Pass 2 builds one scan task per `(instance-file, functionClass)` pair, then merges each result with a pop-then-append. A single source file declaring concrete instances of *two* different functionClasses produces two tasks for that file, and the per-task **pop** meant the second task discarded the first task's instance names — silently dropping that file's instances of every class but the last from the emitted `functionClassInstances` list (which drives the state store/restore code generation).

The pop exists to clear a stale entry left by a *previous run*; it now fires only once per instance file per run (tracked in a `popped` set), so sibling tasks for the same file accumulate. The across-run freshness behavior is unchanged.

**Latency:** verified none of the 198 functionClasses currently shares an instance file with another, so it never fires today — and it's faithful to the Perl original's delete-then-push-per-pair semantics. But it's a real bug for any file that comes to implement instances of two classes.

**Tests:** `scripts/build/tests/test_state_storables_multiclass.py` drives `stateStorables.main()` end-to-end over a fixture tree where one file implements instances of two functionClasses (verified to emit only the *last* class's instance against the buggy code, both against the fix), plus a two-separate-files control confirming the common one-class-per-file case is unchanged. The real-tree `stateStorables.xml` regenerates byte-identical before and after, and the full build-script suite passes (236 tests).

### DIFF
--- a/scripts/build/stateStorables.py
+++ b/scripts/build/stateStorables.py
@@ -235,9 +235,19 @@ def main(argv):
                 p2_tasks.append(
                     (instance_identifier, instance_file, function_class_name)
                 )
+    # The cache entry for a freshly-scanned instance file is popped ONCE (to
+    # discard any stale instances left from a previous run), then every task's
+    # names are appended. A single instance file that implements instances of
+    # more than one functionClass yields one task per class; popping only on
+    # the first of them lets those tasks accumulate, instead of a later task's
+    # pop discarding an earlier one's names (which silently dropped that file's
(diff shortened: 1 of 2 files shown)


abensonca (commented):
Reviewed - looks good.

PR #1230, opened 2026-07-03
perf(build): parallelize sourceDigests' per-file and composite hashing

abensonca:
Stacked on #1229. Tackles the last deferred item from the build-infrastructure review — the one flagged as needing a "find_hash locking rework."

`sourceDigests.py` was the only tree-walking build script with no parallelism, and it runs at every executable link. A **cold** run (first build / after `make clean`) spends ~16s single-threaded: ~2.5s hashing each source file into its `.md5` sidecar, and ~13s walking every type's dependency list to build its `.md5c` composite. A naive parallelization races on the per-file `.md5` sidecars that many types share as dependencies — exactly why this was deferred.

## Approach: one writer per sidecar

Rather than reworking the *locking*, the fix reworks the *structure* so intra-run parallelism is race-free by construction:

- **SourceDigest.py**: the per-file digest computation is extracted verbatim from `find_hash`'s inner loop into `ensure_file_digest` (find_hash calls it — behavior byte-identical).
- **Phase 1** (`_prewarm_file_digests`): compute every per-file `.md5` in the target's dep set in parallel — one task per unique source file, so no two workers write the same `.md5`.
- **Phase 2**: the per-file scan loop (directive extraction + `find_hash` + AST walk) becomes a parallel worker. With `.md5` pre-warmed, each worker only *reads* them and writes only its own file's `.md5c`, so distinct workers never contend on a sidecar. Results merge in source order, preserving the serial run's last-writer-wins semantics.

Locks are forced on for the phase-2 workers only when the scan actually runs in parallel (guarding the rare dependency `.md5` the pre-warm didn't cover, plus concurrent external builds). External-build `LOCKMD5`/`use_locks` semantics are otherwise unchanged, and the serial path (no `make -j`) is byte-for-byte the old behavior.

## Result & verification

**16.1s → 2.3s (~7×) on a cold Galacticus.exe digest run, 20 cores**; warm runs unchanged (~0.8s).

- `Galacticus.md5s.c` is **byte-identical** across serial, parallel, and three repeated cold parallel runs, and matches the pre-change baseline.
- The extraction refactor alone was verified byte-identical before adding any parallelism.
- The pickled blob cache is **deep-equal** serial vs parallel.
- End-to-end: a real `make -j20` relink of `tests.hashes.cryptographic.exe` builds and passes its MD5 self-test.
- Full build-script suite passes (236 tests).

### DIFF
--- a/python/Galacticus/Build/SourceTree/Process/SourceDigest.py
+++ b/python/Galacticus/Build/SourceTree/Process/SourceDigest.py
@@ -131,6 +131,116 @@ def hash_data_files(hasher, files):
                 hasher.update(fh.read())
 
 
+_DATA_FILE_RE = re.compile(
+    r'[\"\'](data/[a-zA-Z0-9_\.\-/]+\.(xml|hdf5))[\"\']'
+)
+_STATIC_DATA_RE = re.compile(
+    r'(char\s*\()??\s*galacticusPath\s*'
+    r'\(\s*pathTypeDataStatic\s*\)\s*\)??\/\/\]\s*'
+    r'[\"\']([a-zA-Z0-9_\.\-/]+\.(xml|hdf5))[\"\']'
+)
+
+
+def ensure_file_digest(source_file_name, suffix, build_path, *, use_locks,
+                       include_files_excluded, report=False):
+    """Return the per-file MD5 digest of `source_file_name`, reading its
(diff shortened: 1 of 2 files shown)


abensonca (commented):
Reviewed - looks good.

PR #1231, opened 2026-07-04
feat(build): library-interface coverage — output arrays, callbacks, and pointer write-back

abensonca:
Extends the libgalacticus wrapper generator with five new argument/return categories, raising method-level readiness from 607 (as first honestly measured) to **622 of 689 (90%)**, and reduces the audit's actionable worklist to a single moot entry. Stacked on #1230.

## Audit restored to truth first

The audit's methods pass had **never worked**: its `<functionClass>` block regex required a bare open tag while every real directive carries attributes, so it reported "0 methods" — and its in-scope bucketing misfired on the literal `class(*)` in blocker-message tails, inflating the worklist ~3×. With those fixed, every subsequent increment was measured against real numbers, and the classification rules stay generator/audit-shared by construction so the two cannot drift.

## New capabilities

| Category | Mechanism | Unblocks |
|---|---|---|
| 1D/2D/logical output arrays (`intent(out\|inout), allocatable`) | `save, target` buffer + `(c_ptr, c_size_t…)` companions, guarded `c_loc` for zero-size/unallocated results, numpy copy (column-major reshape for 2D; kind-narrowing copy for logical) | `wavenumbersLocalMinima`, `construct`, `fitGeneric`, `indicesFromPosition`, `timeLightconeCrossing`, `intervals`, `luminosityTracks`, … |
| Scalar `intent(out)` companions | by-reference dummy; value leads/joins the returned tuple in declaration order | `wavelengths`, `tabulation`, `modelFDF`, … |
| Scalar & dynamic-array returns alongside outputs | function result or subroutine-lowered return companions compose with per-arg companions | `lengthToCellBoundary`, `intervals` |
| Python-callable callbacks (registry-gated) | CFUNCTYPE → `c_funptr` slot in a per-method shim module → Fortran shim matching the Galacticus interface; pointer dummies get a re-aimed procedure-pointer slot | `integrate` (coordinate→Cartesian shim), `integrateOverCrossSection` |
| Complex sized-output buffers | `complex(c_double_complex)` ↔ numpy `complex128`; `dimension(gridCount,…)` extents resolved from the method's own integer inputs — Python pre-allocates, no companions | `windowFunctions` (3D FFT grids) |
| Pointer write-back protocol | `type(X), pointer, intent(out\|inout)` method args cross as by-reference `c_ptr`; guarded `c_f_pointer` in, `c_loc`/null write-back out; Python's `c_void_p` handle updates in place | `next`, `promote`, `control` — **tree iteration from Python**: `while walker.next(node): …` |

## Correctness fixes surfaced along the way

- `mergerTreeWalker.next` / `promote` / `control` were counted ready but shipped **silently broken** — the object pointer crossed by value, so repointing never reached the caller. They were honestly rejected first, then properly supported by the write-back protocol (method-only: constructors have no post-call hook, gated by an `allow_pointer_writeback` flag through the shared predicate).
- `build_fortran_reassignments` re-dispatched on output-array args' declared intrinsics; its logical branch clobbered the output wiring.
- Outbound procedure/`class(*)` pointers, `type(nBodyData)` arrays, and polymorphic-array args are now classified as unsupportable-in-principle/deferred with precise messages instead of sitting in the actionable worklist.
- `Classification.py` was missing from `libgalacticus.Inc`'s Makefile prerequisites.
- Build artifacts (`libgalacticus.so`, `galacticus/`) are now gitignored.

## Verification

- **236 unit/integration tests** (audit regression, classification, generator golden-shape, callbacks, write-back).
- Every category proven end-to-end with **gfortran-16-compiled stubs** driven from the generated Python: exact quadrature through callbacks (∫x dV = 0.5; spherical-shell closed forms), column-major 2D/3D complex indexing, `np.bool_` conversion, save-buffer lifetime across calls, and a 5-node chain walked from Python via the write-back handle.
- **Full library built** (`make GALACTICUS_BUILD_OPTION=lib libgalacticus.so`, gfortran-16) and the interface suite extended and run against it: **130 PASS, 0 FAIL**, including real-library checks of `wavenumbersLocalMinima` (zero-length guard), `integrate` (exact unit-cube/shell integrals), and `integrateOverCrossSection` (exact zero and linearity in σ on a black-body field).

## Remaining

The audit's in-scope worklist is one mixed entry (`timeEvolveTo` — moot beside its unexposable `class(*)` pointer output). Everything else (66 methods) is the deliberately-deferred backlog: internal derived types, non-fc class hierarchies, pointer outputs.


abensonca:
Added `tutorials/` — four executed Jupyter notebooks introducing the library interface (cosmology calculator; matter power spectrum incl. WDM half-mode masses; halo mass functions; Python callbacks with exact analytic checks). Each is committed with output produced against the library built on this branch, and notebooks 2 & 4 exercise the new features (output arrays, both callback flavours).

abensonca:
Follow-up work now on the branch:

* **CI**: the Python-Interface job executes all tutorial notebooks against each freshly built library (`tutorials/executeNotebooks.py`), with the datasets repo checked out for the stellar tutorial.
* **ReadTheDocs**: tutorials render from their committed outputs via nbsphinx (`execute='never'` — CI keeps the outputs truthful); verified with a full local RTD-equivalent build.
* **Two new notebooks**: 05 merger-tree building blocks (M*(z), Parkinson-Cole-Helly branching rates, root-mass sampling — exercising the zero-size output-array path in real use) and 06 stellar populations (IMFs, lifetimes/remnants/yields, SN Ia DTD).
* **Real physics bug found & fixed** while probing for notebook 06: `chabrier2001NumberCumulative` summed its integral limits instead of differencing them, making every cumulative count above 1 M☉ *negative* — this feeds the SN Ia delay-time-distribution normalizations. Fixed, verified against direct numerical integration (ratio 1.000000), and pinned in the interface suite (now 133 PASS).

abensonca:
Survey-geometry tutorial added — and it took two more real bug fixes to get there:

* **FFTW has been silently disabled in every build**: the availability probe compiles `fftw3_config.F90` raw, but its docstring was a bare `!!{RST}` block — an unconditional syntax error, so the probe always registered `FFTW3UNAVAIL` everywhere (CI included). Fixed by comment-prefixing the docstring (`cf74ad142`). CI's next library build should activate FFTW automatically.
* **The full-sky window function was latently broken**: its grid-cell positions omitted `/gridCount`, so the selection function was identically zero. Unobservable until the probe fix made the code path reachable. Fixed and validated to machine precision (2×10⁻¹⁶) against an independent numpy FFT (`bb0f4e610`).
* New `galacticus.verbositySet(level)` library function (`314890633`) — Galacticus's progress bars write straight to fd 1 and land in notebook cell outputs otherwise.
* **Tutorial 07** (`40cc6e53c`): SDSS mask footprint reconstructed from `pointIncluded` (Monte-Carlo fraction asserted against the mask's solid angle, which matches mangle's harmonize value to 12 digits), Vmax(M), and FFTW-computed 3D window functions validated against numpy and the analytic top-hat — all in executed cells. Suite now 134 PASS; all seven notebooks green via the CI runner.

Note for CI: tutorial 07's first run downloads the SDSS mangle mask (~zenodo) and parses 200k polygons (~1–4 min) — if that proves flaky we can add an actions/cache for `datasets/dynamic/surveyGeometry`.

abensonca:
Real-survey window functions are now live from Python, plus the P(k)-convolution finale:

* **Latent parser bug fixed** (`c77744e88`): `windowRead` read mangle polygon headers by fixed word position — for *pixelized* files (all the modern res6d masks) the "solid angle" it stored was actually the **pixel number**. The largest-polygon-first search ordering has been garbage for those masks all along; surfaced when the new sampler aborted on zero-area "polygons". Now parsed by field label, handling both layouts.
* **Mangle window functions implemented** (`7b6c864b1`): previously an `Error_Report` stub. Monte-Carlo sampling with an O(1)-per-sample polygon sampler (∝ solid angle — exactly uniform for balkanized masks — with a bounded-bias sliver fallback), depths uniform in volume, FFTW'd. `surveyGeometryMangle` gains a `randomNumberGenerator_` component; Bernardi+13's constructors accept it; `windowFunctionAvailable()` gates on it (pinned in the suite — 136 PASS). Validated from Python: 289/300 of the reconstructed selection function's most-occupied cells confirmed by the independent `pointIncluded` path, zero false interiors.
* **Tutorial 07 finale** (`57d595314`): the SDSS window side-by-side with the full-sky one (the footprint's anisotropic imprint on Fourier space), and Tutorial 2's linear P(k) convolved with |W|² via the convolution theorem — obs/true asserted at honest few-per-cent bounds for this compact kernel. All seven notebooks green (~140 s for 07).

abensonca:
RNG constructor argument rolled out to all remaining mangle surveys (`4f8bde64b`): Baldry+12 GAMA, Davidzon+13 VIPERS, Muzzin+13 ULTRAVISTA, Moustakas+13 PRIMUS, Tomczak+14 ZFOURGE, and both Local Group geometries (which also gained the destructors they were missing). Every mangle survey can now compute Monte-Carlo window functions from Python and from parameter files. Availability gating (False without an RNG, True with) verified live for all eight surveys and pinned in the interface suite — now **150 PASS**; all seven notebooks re-executed green.

abensonca (commented):
Reviewed - all looks good.

PR #1234, opened 2026-07-05
fix(build): macOS include case-collision and .fl staleness hardening

abensonca:
Two loose ends from the macOS CI investigation (stacked on #1230, merge after it).

**1. `libgalacticus.inc`/`.Inc` case collision (`df5f8a5e2`)**

On case-insensitive filesystems (macOS APFS) the processed include `libgalacticus.inc` and the generated interface `libgalacticus.Inc` are the same file. Consequences: in mixed executable-then-library build trees the executable build's placeholder masqueraded as an up-to-date `libgalacticus.Inc`, silently skipping `libraryInterfaces.py` and building the library against an empty interface; on incremental library rebuilds the processed include overwrote the raw interface in place, reprocessing its own cpp output under a perpetually-bumped timestamp. (Fresh single-mode CI trees were unaffected, which is why `Build-Library-MacOS` always passed.)

- Processed include renamed to `libgalacticus.preprocessed.inc` — no case twin.
- Non-library builds get an explicit, documented empty placeholder. Previously the generic `%.inc ← %.Inc ← %.Inc.up` pattern chain manufactured a husk by running `sed` on a nonexistent file with the failure masked mid-pipeline (the `sed: … No such file or directory` noise visible in every historical build log).
- The generic `%.inc` recipe now fails loudly on a missing input instead of writing a husk of cpp boilerplate.

**2. `.fl` sidecar staleness (`e1a91530b`)**

The `.fl` library sidecars were populated solely by prerequisite-less echo rules: once a `.fl` existed its recipe could never run again, so a changed library dependency (new/removed `use` of a library-mapped module) or a corrupted `.fl` was never repaired short of a clean build — the link silently kept the stale library set. `useDependencies.py` now writes each `.fl` directly during the scan via the only-if-changed helper (mtimes preserved when unchanged, so no spurious rebuilds); the echo rules remain as the recovery path for manually-deleted files.

**Verification**: `make -n` rule-selection checks in both exe and lib modes; loud-guard failure demonstrated; on a warm tree all pre-existing `.fl` byte-identical with mtimes preserved; a deliberately corrupted `.fl` repaired on the next scan; emitted `Makefile_Use_Dependencies` byte-identical old-code vs new-code; 662 build-script tests pass. The macOS-specific behavior (mixed-mode tree correctness) should be confirmed on a Mac: `make Galacticus.exe && make GALACTICUS_BUILD_OPTION=lib libgalacticus.so` in one tree, then `nm libgalacticus.so | grep -ci galacticus` — nonzero proves the interface is real.


abensonca:
Third commit added: `efbef1f5e` extends the case-collision fix to the **general** `.Inc→.inc` convention — every BUILDPATH preprocessor intermediate is now named `%.p.Inc` (matching the libgalacticus chain), so no include pair places two case-twin names in BUILDPATH. Processed `.inc` names are unchanged (no Fortran or generated-code changes). On macOS this ends the perpetual re-preprocessing of every include pair on incremental builds. Verified: byte-identical processed output (modulo line markers naming the new intermediate), generated rules carry renamed targets with the components generator's Python prerequisites intact, loud missing-input guard works through the new chain, consumer syntax-only compile resolves, 662 tests pass.

PR #1236, opened 2026-07-06
Merger trees from Python: opaque tree handles, per-node property extraction, and a tutorial

abensonca:
Stacked on #1231. Completes the tree-access arc of the Python library interface: merger trees can now be **built, walked, and read entirely from Python**.

## Library capabilities

* **Opaque shared-type pointer returns** — methods returning `type(mergerTree), pointer` (any type in `_SHARED_TYPE_MODULES`) now cross bind(c) as `c_loc` handles (`c_null_ptr` when disassociated → `None` in Python). Ownership stays with Fortran. Unblocks `mergerTreeConstructor.construct`; readiness rises 622 → 624 (90%).
* **`extractScalar` on `nodePropertyExtractor`** — a base-class `<method>` dispatching to the scalar subclass's `extract`, giving library users per-node property access through the walker loop.
* **Library initialization** — `libGalacticusInitL` now calls `Functions_Global_Set()` (mirroring `Galacticus.exe` startup; without it any code path crossing a deferred-binding seam aborted), and a new `galacticus.nodesInitialize()` boots the node-component class system (idempotent, default component selections) before any node or tree is created.
* **Fix: `intent(out)` logical scalars now copy back** — the wrapper kind-widened `logical(c_bool)` args into a default-kind local but never narrowed the result back after the call, so e.g. `construct`'s `finished` flag stayed `False` forever (an infinite suite loop ending in a null-handle dereference). Pinned in the suite: exhausting the tree suite now flips `finished` and returns `None`.

## Tutorial 8: merger trees

The payoff notebook ([executed output committed](tutorials/08-merger-trees.ipynb), ~3 s):

```python
finished = ctypes.c_bool(False)
tree     = treeConstructor.construct(1, finished)
walker   = galacticus.mergerTreeWalkerIsolatedNodes(tree)
node     = ctypes.c_void_p(None)
while walker.next(node):
    masses.append(extractorMass.extractScalar(node))
```

Cole et al. (2000) trees from a hand-assembled 12-object graph: the assembly wedge of all 1207 nodes of one tree, main-branch-proxy assembly histories across the 8-tree suite, and per-node virial scalings via `darkMatterHaloScale`. Wired into the docs toctree, tutorials README, and the CI notebook-execution job.

## Testing

* `testSuite/test-Python-interface.py`: new merger-tree build/walk/extract section — **158 checks pass**. (Also loosens the δ_c / dn/dlnM pins to rtol=1e-4: they interpolate the spherical-collapse tabulation, whose cached grid layout depends on query history — the tree machinery tabulates δ_c to z=10⁵, shifting δ_c(t≈13.8) by ~4e-5; same rationale as the existing Δ_vir pin.)
* All 8 tutorial notebooks execute cleanly against the rebuilt library.
* Interface audit: Ready 624/689 (90%), in-scope blocked 1.


abensonca (commented):
Reviewed - looks good.

Issue #75, opened 2020-03-05
Coordinates/Vectors/Etc unification


abensonca:
Currently we have a `coordinates` class which supports spherical, cylindrical, and Cartesian coordinates. We also have a `vector` class used for linear algebra work. And in some functions position arguments are specified just as a 3-element double precision array, with the coordinate system implicitly assumed.

`coordinates` class is equivalent to a vector, except that it is specific to a coordinate system (which is useful) and has automatic conversion between coordinate systems. 

We should unify these approaches:

- Remove the use of 3-element arrays as a means of specifying positions;
- Create assignment operators for `coordinates` to `vector` class (`vector` class always stores Cartesian components, so just need to convert to Cartesian first, which is trivial);

Should `coordinates` even just be a subclass of `vector`, allowing us to perform any vector operations directly on a `coordinates` object?

abensonca:
And `vector` maybe just becomes a member of the `tensor` class.

abensonca:
# Issue #75 review: Coordinates/Vectors/Etc unification

*Codebase review, 2026-07-05. Issue filed 2020-03-05; the relevant classes are essentially unchanged since then (only docs/copyright/memory-leak commits), so the issue's landscape is intact — but one of its premises needs revising (see §3.2).*

## 1. The current landscape

There are **six** coexisting representations of a 3-vector/position in the codebase:

| Representation | Location | Files using | Character |
|---|---|---|---|
| `coordinate` hierarchy (`coordinateCartesian`/`Spherical`/`Cylindrical`) | `source/objects/coordinates.F90` | 152 | Value type, `position(3)` payload, system-aware, conversion via `assignment(=)`. The interface currency of the `massDistribution`/kinematics/computational-domain/radiative-transfer families. |
| `vector` + `matrix` (GSL-backed) | `source/math/linear_algebra.F90` | 34 | Runtime-dimension, heap-allocated, reference-counted. 2–3 heap allocations + memcpy **per operation**. |
| `Vectors` free functions on `dimension(3)` arrays | `source/math/vector.F90` | 45 | `Vector_Magnitude` (43 files), `Vector_Product` (12), outer products (BLAS). Stateless, allocation-free. |
[...]

PR #1259, opened 2026-07-13
refactor: coordinates cleanup (#75 Phase A) — remove dead Coordinate_Systems, correct scale() docs

abensonca:
Part of #75 (coordinates/vectors/etc. unification). This is **Phase A** of a phased, independently-mergeable plan: dead-code and stale-workaround cleanup. Low risk, no behavior change.

## What changed

**1. Delete the dead `Coordinate_Systems` module**
- `source/geometry/coordinate_systems.F90` and `source/tests/geometry/coordinate_systems.F90` are removed. The module duplicated conversion mathematics already implemented in the `coordinate` class and had **no production callers** (only its own test used it).
- Its 8 hand-verified conversion points are **preserved** in a new `source/tests/coordinates.F90` (`tests.coordinates.exe`) that exercises the `coordinate` class conversions directly (cartesian↔spherical↔cylindrical via assignment). The CI matrix is updated to run the new test in place of the old one.

**2. Correct the stale `scale()` "workaround" documentation**
- The three PR-37336 gfortran-finalization `<workaround>` directives on `Coordinates_*_Scale` are removed. That leak is fixed in gfortran ≥ 13 (we require gfortran 16).
- They're replaced with an accurate note. Investigation showed the original premise was stale: `scale()` is **not** merely a finalization workaround — it is the only way to scale into an *unallocated* `class(coordinate), allocatable` target. The overloaded `*`/`/` operators return function results and so require an already-allocated (typically concrete-typed) assignment target; using them on an unallocated polymorphic target selects the defined assignment on the abstract-typed LHS and fails at runtime (verified with a probe). `scale()` is therefore retained, and the `*`/`/` operators are **kept** because they have real users (e.g. `Enzo_hydrostatic`).

### Note on the original plan
The Phase A plan proposed deleting `scale()` in favour of the operators and adding concrete-typed operators. That turned out to rest on stale premises (much as the plan itself flagged for its §3.2): `scale()` cannot be removed without regressing/breaking the polymorphic scaler decorators, and the operators are not dead. This PR does the defensible cleanup that the phase was actually after.

## Verification
- Build: clean (gfortran 16, `-O3 -flto`).
- `tests.coordinates.exe`: 32/32 pass.
- `tests.mass_distributions.exe`: 148/148 pass (exercises `scale()` in the spherical/cylindrical scaler decorators **and** the `*`/`/` operators in `Enzo_hydrostatic`'s density loop).
- valgrind on `tests.mass_distributions`: **no coordinate-related leaks**. The finalization fix is confirmed on real code — no per-call leak from the operator hot path. (The 642 bytes of pre-existing baseline loss are all in unrelated setup code, e.g. the tabulated-profile file reader.)


abensonca (commented):
Reviewed - looks good.

PR #1260, opened 2026-07-13
refactor: type-symmetric massDistribution interface (#75 Phase B)

abensonca:
Part of #75 (coordinates/vectors/etc. unification). **Phase B** makes the `massDistribution` family interface type-symmetric: the methods that took `class(coordinate)` inputs but returned raw `double precision, dimension(3)` arrays now also **return** a typed `type(coordinateCartesian)` object. This removes the wrap/unwrap asymmetry the issue calls out and fixes a latent physics footgun (P4).

## What changed

**Typed returns** (via the `_class.F90` `<method>` directives):
- `acceleration` → `type(coordinateCartesian)`
- `positionSample` → `type(coordinateCartesian)`
- `chandrasekharIntegral` → `type(coordinateCartesian)`

**P4 velocity footgun fixed** — `chandrasekharIntegral`'s `velocity` argument is retyped `class(coordinate)` → `type(coordinateCartesian)` (issue option (a)). Previously a velocity expressed in a non-Cartesian basis would be silently transformed with the *point* transformation (wrong for a velocity); only Cartesian velocities were ever actually passed, so this makes the type system honest without changing behavior.

**Implementations** — all 15 (spherical/cylindrical `_class`, composite, spherical & cylindrical scalers, Gaussian ellipsoid, exponential disk, isothermal) build the result and assign the typed return.

**Callers** — updated only where they do *arithmetic* on the result (`Chandrasekhar1943`, the `distributed` radiative-transfer source, and the `mass_distributions` test). Callers that merely assign the result into a `dimension(3)` variable are **unchanged** — the existing array↔coordinate assignment overloads bridge them transparently (e.g. `potentialDifferenceIntegrand`, `satellite_orbits`).

**Supporting changes:**
- `coordinates.F90`: the base `coordinate%position` is default-initialized to zero, so the argument-less `coordinateCartesian()` structure constructor (emitted by auto-generated `functionClass` code for typed returns) is valid. This also eliminates a class of uninitialized-coordinate reads.
- `cylindricalChandrasekharIntegral`: the result is initialized to zero up front, so the velocity-dispersion early-return paths now yield a well-defined **zero** integral (they previously returned an uninitialized array — a latent bug).

## Verification
- Full **`Galacticus.exe` builds** (gfortran 16, `-O3 -flto`) — compiles every caller across the tree, including the dynamical-friction, radiative-transfer, and node-operator consumers.
- **`tests.mass_distributions.exe`: 148/148 pass** (exercises all acceleration/positionSample paths and both scaler decorators).
- **valgrind**: 642 bytes / 45 blocks definitely lost — byte-for-byte identical to the pre-existing baseline (all in the unrelated tabulated-profile file reader). The typed returns are stack value types and add **zero** heap traffic.

## Depends on / relation to Phase A
Independent of #1259 (Phase A); branched from `master`. No file overlap beyond both touching `coordinates.F90` in non-conflicting regions.

### DIFF
--- a/source/mass_distributions/Gaussian_ellipsoid.F90
+++ b/source/mass_distributions/Gaussian_ellipsoid.F90
@@ -336,16 +336,16 @@ contains
     use :: Linear_Algebra                  , only : assignment(=)                 , operator(*)        , vector
     use :: Numerical_Constants_Astronomical, only : gravitationalConstant_internal
     implicit none
-    double precision                                   , dimension(3)  :: gaussianEllipsoidAcceleration
+    type            (coordinateCartesian              )                :: gaussianEllipsoidAcceleration
     class           (massDistributionGaussianEllipsoid), intent(inout) :: self
     class           (coordinate                       ), intent(in   ) :: coordinates
     type            (coordinateCartesian              )                :: coordinatesCartesian
     double precision                                   , dimension(3)  :: positionCartesian            , positionCartesianScaleFree , &
-         &                                                                accelerationScaleFree
+         &                                                                accelerationScaleFree        , accelerationArray
     integer                                                            :: i
     type            (vector                            )               :: positionVector               , positionVectorUnrotated    , &
          &                                                                accelerationVector           , accelerationVectorUnrotated
-    
+
     ! Ensure that acceleration is tabulated.
     call self%accelerationTabulate()
     ! Construct the scale-free (and rotated) position.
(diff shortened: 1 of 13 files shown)


abensonca:
## Performance verification (darkMatterOnlySubHalos)

Profiled against the nightly benchmarking harness's `darkMatterOnlySubHalos` model (8 P-cores, `OMP_NUM_THREADS=8`, `perf stat -d -d -d`), comparing this branch to its exact base commit `abc87c1cb` built and run in the same environment. Retired **instructions** is the reported metric (deterministic to ~0.05%, unlike wall time which carries ±2–4 s machine noise).

| Build | Instructions | Δ vs base |
|---|---|---|
| base `abc87c1cb` | 23.6064e12 | — |
| typed returns, initial | 23.9542e12 | **+1.47%** |
| typed returns + null constructor (this branch) | 23.6059e12 | **−0.003% (within noise)** |

The initial typed-return commit gave the base `coordinate` type a `position = 0.0d0` default initializer (needed so the auto-generated `coordinateCartesian()` result construction compiles). That zero-init is applied to *every* coordinate construction and, because coordinate assignment is a defined assignment with an `intent(out)` argument, is not elided by the compiler — costing +1.47% instructions in this coordinate-heavy model. Commit 2e72e6463 replaces it with an explicit null constructor so only the generated result init pays for the zeroing. **Net effect on this branch: no measurable regression** (instructions identical to base), `tests.mass_distributions` 148/148.

PR #1261, opened 2026-07-14
perf: de-spatialize Linear_Algebra in Gaussian ellipsoid hot paths (#75 Phase C)

abensonca:
Part of #75 (coordinates/vectors/etc. unification). **Phase C** removes the GSL-backed `Linear_Algebra` `vector`/`matrix` usage from the Gaussian ellipsoid's per-evaluation rotations (issue review item P1 / plan item 7).

> **Stacked on #1260 (Phase B).** This PR targets the Phase B branch, so the diff shown is Phase C only. Rebase/retarget to `master` once Phase B merges. (It stacks because both phases touch `gaussianEllipsoidAcceleration`.)

## What changed

The Gaussian ellipsoid rotates positions into — and accelerations out of — the frame aligned with its principal axes on **every** density and acceleration evaluation. That was done with the GSL-backed `vector`/`matrix` types: each call heap-allocated `vector` objects, did a `matrix*vector` product with GSL memcpys, and freed them (finalizer/reference-count churn) — all inside quadrature integrands and, via `cylindricalChandrasekharIntegral` (disk dynamical friction), inside the satellite-orbit ODE loop.

- `rotationIn`/`rotationOut` are now stored as plain `double precision, dimension(3,3)` arrays instead of GSL `type(matrix)`.
- They're still computed once in `initialize()` using the GSL matrix type (`matrixRotation` + `inverse()`, not a hot path) and then extracted to arrays.
- The density/acceleration hot paths apply them with `matmul` — no heap allocation, no GSL calls.
- The `rotation` constructor argument is left as a GSL `matrix`, so callers (`cylindrical/_class.F90`) are unchanged.

## Performance (milkyWay validation model)

Profiled Phase C against its exact base (Phase B tip) built and run back-to-back in the same environment (8 P-cores, `OMP_NUM_THREADS=8`, `perf stat`):

| Build | Instructions | Wall (s) |
|---|---|---|
| Phase B tip | 24.538e12 | 373.3 |
| **Phase C** | **23.285e12** | **346.8** |
| **Delta** | **−5.1%** | **−7.1%** |

Phase B tip matches the nightly milkyWay band (24.5–24.7e12), so the **−5.1% retired instructions is attributable to this change.** It's this large because the ellipsoid `acceleration` is evaluated per step of the satellite-orbit ODE integration (disk dynamical friction), so removing the per-call GSL vector round-trips compounds. (The darkMatterOnlySubHalos model barely uses the ellipsoid path and is unaffected.)

## Correctness
`matmul` produces identical results to the GSL `matrix*vector`: `tests.mass_distributions` **148/148**, including the "Acceleration of rotated ellipsoid" test. Full `Galacticus.exe` builds.

## Follow-up (not in this PR)
`cylindricalChandrasekharIntegral` re-`initialize()`s the ellipsoid velocity distribution every call, which still builds a GSL rotation matrix + inverse per call. Hoisting/caching that is a further opportunity beyond this phase's scope.

### DIFF
--- a/source/mass_distributions/Gaussian_ellipsoid.F90
+++ b/source/mass_distributions/Gaussian_ellipsoid.F90
@@ -34,30 +34,30 @@
      !!{RST
      A Gaussian ellipsoid mass distribution. The formulation and calculations follow the parameterizations and approach of :cite:t:`chandrasekhar_ellipsoidal_1987`. The ellipsoidal has scale lengths :math:`a_\mathrm{i}` along each of the three perpendicular axes (which are assumed to be aligned with the Cartesian :math:`(x,y,z)` axes).
      !!}
-     double precision                     , dimension(3          ) :: scaleLength
-     double precision                                              :: mass                             , density_
-     logical                                                       :: accelerationInitialized
-     double precision        , allocatable, dimension(:          ) :: accelerationX                    , accelerationScaleLength
-     double precision        , allocatable, dimension(:,:,:,:,:,:) :: accelerationVector
-     double precision                                              :: accelerationXMinimumLog          , accelerationXMaximumLog               , &
-          &                                                           accelerationScaleLengthMinimumLog, accelerationScaleLengthMaximumLog     , &
-          &                                                           accelerationXInverseInterval     , accelerationScaleLengthInverseInterval, &
-          &                                                           scaleLengthMaximum
-     integer                              , dimension(3          ) :: axesMapIn                        , axesMapOut
-     double precision                     , dimension(2          ) :: axisRatio
+     double precision             , dimension(3          ) :: scaleLength
+     double precision                                      :: mass                             , density_
+     logical                                               :: accelerationInitialized
+     double precision, allocatable, dimension(:          ) :: accelerationX                    , accelerationScaleLength
+     double precision, allocatable, dimension(:,:,:,:,:,:) :: accelerationVector
+     double precision                                      :: accelerationXMinimumLog          , accelerationXMaximumLog               , &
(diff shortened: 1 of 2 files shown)


abensonca:
## Follow-up commit: remove remaining GSL from the cylindrical Chandrasekhar path

Commit 8c788c58f extends this to `cylindricalChandrasekharIntegral`, which re-`initialize()`s the ellipsoid velocity distribution every call (noted as follow-up in the PR description). The per-call rotation matrix is now a plain `dimension(3,3)` array (no GSL matrix construction), and the reverse rotation is computed as `transpose()` of the orthonormal rotation rather than a GSL LU inversion. `cylindrical/_class.F90` no longer uses `Linear_Algebra` at all.

Cumulative milkyWay profiling (each built + run back-to-back, same machine, `perf stat`):

| Build | Instructions | Wall (s) |
|---|---|---|
| Phase B tip (base) | 24.538e12 | 373.3 |
| + matmul rotations (first commit) | 23.285e12 | 346.8 |
| + cylindrical de-GSL (this commit) | **22.864e12** | **339.7** |
| **Total vs base** | **−6.8%** | **−9.0%** |

Retired instructions decrease monotonically; the whole branch removes ~6.8% of instructions from the milkyWay model by eliminating GSL vector/matrix heap allocation from the ellipsoid density/acceleration rotations and the disk-dynamical-friction rotation setup. Correctness unchanged: `tests.mass_distributions` 148/148, including the rotated-ellipsoid acceleration test (which now exercises the `transpose`-based inverse). Full `Galacticus.exe` builds.

PR #1263, opened 2026-07-14
Add Lancaster2020 dynamical friction implementation

yzhaoastro:
## Summary

- Added an implementation of the Lancaster et al. (2020) dynamical friction model.
- Added a wrapper dynamical friction class that applies a multiplicative scale factor to the acceleration returned by another dynamical friction model.
- Fixed an issue in DMO halo evolution where an initially NFW halo could incorrectly transition to a soliton+NFW profile during subsequent evolution. The initial halo profile is now cached and used consistently throughout the evolution.

## Type of change

- [x] Bug fix
- [x] New feature
- [x] Docs / tooling
- [ ] Other

## How to test

- Built Galacticus successfully.
- Verified that the Lancaster (2020) dynamical friction model produces the expected acceleration.
- Verified that the scaling wrapper correctly rescales the child dynamical friction acceleration.
- Verified that DMO halos remain consistently NFW or soliton+NFW according to their initial classification.

## Checklist

- [x] I ran relevant tests (or explain why not):
- [x] If this PR is from a fork: I enabled 'Allow edits from maintainers'


abensonca (approved):
Everything looks good. 

For the new `solitonStatus` my understanding is that this is initially `0` which causes it to attempt to evaluate the soliton solution. If a solution is found, it is then set to `+1` (in which case on all subsequent attempts it also tries to find a soliton solution). If no solution is found, it is set to `-1`, and it is treated as NFW in all subsequent attempts. Is that correct?

One concern I have about this is that if we ever fail to find a soliton solution, we _never_ attempt to find one again. That does solve the original problem, but I wonder if it causes a new one - if we fail to find a soliton solution at one timestep, then we are fixed at NFW forever, even if at a later timestep a soliton solution becomes possible again. I'm not sure that there is a better solution - I just wanted to mention this so that we can think about it further.

I also changed the `solitonStatus` meta-property to integer type - not really necessary but more appropriate for an indicator value such as this.

Other than that I just made some minor formatting fixes. I'll merge this once the tests are all passed.

abensonca:
:warning: This work adds new dynamical friction classes. PR #1260 also touched dynamical friction classes - verify that the new classes here are consistent with that before merging. :warning:

abensonca:
> ⚠️ This work adds new dynamical friction classes. PR #1260 also touched dynamical friction classes - verify that the new classes here are consistent with that before merging. ⚠️

Verified - looks like there should be no conflicts.

yzhaoastro:
Yes, your understanding is correct. I had the same concern about permanently switching to NFW after a single failed attempt.

One possible alternative, which Xiaolong suggested before, is that if we fail to find a soliton solution, we could use a smoothly truncated NFW profile as the fallback instead of a standard NFW profile. The idea is that, if halos with this issue are only a small fraction of the total population, truncating the NFW profile at around ( \lambda_{\rm dB} ) (preferably with a smooth transition) would be more physically motivated, since we expect the inner region below that scale to be supported by quantum pressure and therefore be cored.

This might also avoid the issue of permanently classifying a halo as NFW after a single failed solve, while still providing a simple and robust fallback when the solver does not converge. I'm not sure whether this would be a better overall solution, but it may be worth considering.

abensonca:
> One possible alternative, which Xiaolong suggested before, is that if we fail to find a soliton solution, we could use a smoothly truncated NFW profile as the fallback instead of a standard NFW profile. The idea is that, if halos with this issue are only a small fraction of the total population, truncating the NFW profile at around ( \lambda_{\rm dB} ) (preferably with a smooth transition) would be more physically motivated, since we expect the inner region below that scale to be supported by quantum pressure and therefore be cored.
> 
> This might also avoid the issue of permanently classifying a halo as NFW after a single failed solve, while still providing a simple and robust fallback when the solver does not converge. I'm not sure whether this would be a better overall solution, but it may be worth considering.

That's definitely worth considering. Let's keep this in mind, pending the results of your analysis of the core mass-halo mass relation (which might change the behavior around soliton radius finding anyway).

PR #1313, opened 2026-08-01
Share radius-specifier evaluation, and allow radii as fractions of soliton radii

abensonca:
## Summary

Two pieces of work on `Galactic_Structure_Radii_Definitions`.

**Deduplicate radius-specifier evaluation.** Every consumer of the radius-definition decoder carried its own copy of the logic turning a decoded specifier into an actual radius in a `treeNode` — a `select case` over all fourteen radius types, a set of `*IsNeeded` flags, and a preamble gathering the components those types need. That was ~600 lines duplicated across eight property extractors, and it had drifted between copies. It is now shared, via a `radiusDefinitions` type holding the decoded specifiers plus what they require, and a `radiusResolver` which evaluates them in one node.

**Allow radii as fractions of fuzzy-dark-matter soliton radii.** New `solitonRadiusCore` and `solitonRadiusSoliton` radius types, read from the meta-properties the soliton profiles create. Unlike every other radius type these can be *undefined*, and are reported with a sentinel rather than as zero.

## The undefined-radius problem

A soliton radius has no value when the node carries only the generic dark matter profile component, when no class creates the meta-properties at all (any model without a soliton-forming profile), or when the halo simply grew no soliton — the soliton profiles signal that last case by storing `-1`.

Reporting these as a radius of zero would be actively harmful: a soliton model contains a profile which diverges at the center, so evaluating the density there raises a floating-point exception. They are instead reported as `radiusUndefined = -huge(0.0d0)`. A large negative finite value is used because no physical radius is negative, and because `-ffinite-math-only` rules out `NaN` and infinity. Since `-ffpe-trap=overflow` is on unconditionally, multiplying the sentinel would itself trap — so every consumer tests for it *before* performing any arithmetic on the radius, and the resolver propagates it rather than multiplying by it.

Whether the meta-properties have a creator is tested when radii are *evaluated*, not when they are decoded: the order in which classes are constructed while parsing parameters is undefined, so the creator flag is not reliably set at decode time.

This needed a new query in the build system. Registering a meta-property with `isCreator="no"` yields a valid ID, but reading one which nothing creates is fatal by design (`metaPropertyNoCreator`). There was no way to ask the question without triggering that, so `<class><Label>Rank<N>MetaPropertyCrtd` is now generated, bound as `<label>Rank<N>MetaPropertyIsCreated`. The internal name is abbreviated for the same reason as `...MetaPropertyRef`: spelled out it would reach 67 characters, over the Fortran limit of 63.

## Two behavior changes, called out

- **`galacticMassFraction` now uses the galactic mass** (`fd305d2d`). It is documented as the radius enclosing a fraction of the *galactic* mass, but seven of the eight extractors located it using the stellar mass, making it identical to `stellarMassFraction`. Only `velocityDispersion` did it correctly. This changes output for anyone using that specifier. Isolated in its own commit.
- **The velocity dispersion guard now tests the integration range** (`9efc0c88`). The old guard enumerated radius types whose component might have zero extent; what it actually protects is the line-of-sight integrations, whose outer limit collapses to zero in those cases. Testing `radiusOuter_ <= 0` directly is equivalent for the listed types and also covers ones it missed — an absolute radius of zero, and a radius enclosing a fraction of a mass which is itself zero.

## Verification

The refactor is **bit-for-bit identical**: a model exercising all fourteen radius types across the eight affected extractors produces 69 datasets agreeing at zero tolerance against a build of the base commit. Both models were run with `OMP_NUM_THREADS=1`.

After the `galacticMassFraction` fix, the only differences are the two `galacticMassFraction` columns, plus round-off residuals (6e-11, 7e-9) in extractors that were not touched. Those residuals are the 20-slot memoized `massDistribution` cache in `treeNode`: changing which distributions are requested changes cache occupancy, and rebuilt `adiabaticGnedin2004` distributions reconverge marginally differently. Confirmed by re-running with only the changed specifiers removed, which restores bit-identity everywhere else.

Tests: `test-outputs`, `test-reproducibility` (13/13, including both `rotationCurve` assertions against stored values), `test-massProfileSatelliteBoundMass`, `test-soliton-coremass`, `test-tidallyTruncatedNFWFit`, and the new `test-radius-specifiers-soliton` (11/11) all pass, as do 835 `pytest` tests.

**Not covered numerically:** `nuclearStarCluster*` radius types and `galacticLightFraction` — neither is reachable from the parameter files used, needing an NSC component and a luminosity setup respectively. Both are covered by compilation and inspection only. Separately, because `densityProfile`, `projectedDensity`, `projectedMass` and `rotationCurve` crash on `r=0` today, the full ladder was exercised through `massProfile` (which tolerates it); since the resolver is shared this tests every branch of the radius computation, but not every extractor over every type.

## Known, out of scope

A radius which is legitimately *zero* rather than undefined — say `1.0` disk radii on a node with no disk — still crashes consumers that cannot evaluate there, with an opaque FPE (`nfwDensity`) or divide-by-zero (`projectedMass`, from `log(0)`). 40 of 175 resolved radii in the test model were zero, so this is routine rather than exotic. It is arguably user error, but deserves a comprehensible error message. Filed as #1314.

Note that the velocity dispersion change in this PR incidentally fixes one of those cases: `velocityDispersion` no longer traps at a zero radius arising from a mass fraction.


abensonca (commented):
Looks good - just missing `<methods>` blocks for two new classes - will add those.

PR #1315, opened 2026-08-01
Clean up LaTeX left over from the RST documentation migration

abensonca:
## Summary

An audit of the tree for LaTeX left over from the June 2026 RST migration
(`db11f0680`), and the cleanup it turned up. The repo now contains no LaTeX
outside the two islands where it is still correct by design, and a CI check
keeps it that way.

**231 residual constructs across `source/`, and 156 tokens leaking into the
rendered ReadTheDocs pages, are now zero.**

## Why CI did not catch this

`convertDocstringsToRST.py --check`, which `Validate-Docstrings-RST` runs,
verifies that every docstring has *been converted*. It cannot verify one came
out clean: the conversion is idempotent, so it skips any region already marked
`!!{RST` / `docformat="rst"`. Every finding here sat inside an already-marked
region, so `--check` exited 0 throughout.

The last commit adds `--check-residual`, which scans the marked regions
themselves. It ships with an empty baseline.

## What was found

| | |
|---|---|
| `\AA`, `{\normalfont \ttfamily …}`, `$…$`, `\citep`, `\href`, `\newline` | ordinary residue, mechanically converted |
| `\_ \% \#` escapes | 54, including the `\%2F` sequences that were corrupting the WolframAlpha URLs in `beta_profile.F90` |
| LaTeX inter-word `\ ` | **103** — see below |
| `\argin` `\argout` `\logicalzero` `\textcolor{red}{…}` | 26, macros whose definitions no longer exist anywhere |
| `\mono{}` in the code generators | **157**, in two different string forms |
| `!#` directive syntax in the developer guide | 18 examples, 103 lines, of a syntax used nowhere in the tree |

### The one that silently corrupts prose

LaTeX's inter-word `\ ` (`e.g.\ virial radius`) is, in RST, an *escaped* space
— docutils **deletes** it, so the page reads "e.g.virial radius". 103 of these
existed. The same `\ ` is legitimate and used 1266 times where it abuts inline
markup (``` ``literal``\ s ```, `` cm\ :math:`^2` ``), so both the fix and the
new check are adjacency-aware.

### Macros with no surviving definition

`\argin` / `\logicalzero` and friends could not be expanded by the migration,
so its brace-stripping pass fused them into the text (`\textcolor{red}{…}` →
`\textcolorred…`). Their meanings were recovered from `doc/Galacticus.tex` as
it stood before deletion — a red `<type>` for the argument type, a blue arrow
for direction — and rewritten as an inline-literal type plus an explicit
`[in]`/`[out]` marker.

### The guide was breaking contributors

`creating-a-new-class.rst` taught bare `!!{` in 22 examples. A file written by
following it verbatim fails `Validate-Docstrings-RST` — confirmed by running
the check on one. Both that page and `development.rst` now explain what the
`RST` marker is for.

## Left alone, deliberately

`<entry description="…">` and `<constant …/>` attributes stay LaTeX —
`extractDocsRST` runs them through `latex_to_rst` at build time, which is why
`\gls{}` survives there. The new check excludes them by construction rather
than by exception list.

The `returnType` / `arguments` metadata in the component builder is converted
but currently reaches no output; its only consumer was the `codeAnalyzer.py`
removed here. Wiring it into the RST method docs is filed separately.

## Verification

- `extractDocsRST` output re-parsed with docutils, before vs after: 56 messages
  both ways, **no page gains one**
- XML well-formedness of every changed directive block compared against
  `master`: **no new breakage**
- Generator changes verified by running `scripts/build/preprocess.py` and
  diffing the generated Fortran — only the intended changes, all inside `!<`
  comments, so compilation is untouched. (The unit tests do not assert on these
  markers.)
- `pytest`: **853 passed, 1 skipped**; 22 new tests pin the checker's silences
  as much as its detections
- Both `--check` and `--check-residual` pass

## Review notes

- Commits are separable and are best read in order; the 84-file first commit is
  mechanical, the interesting judgement is in commits 2 and 3.
- The `[in]`/`[out]` convention replacing the blue arrows is a choice, not a
  recovery — the arrows' rendering is gone and something had to be picked.
- The two `tracks/file.F90` methods document only their arguments, with no
  prose saying what they do. I left that gap rather than invent text; it wants
  someone who knows the interpolation code.
- A local `make Galacticus.exe` was still running when this was opened.


abensonca:
Local build and smoke test now done, closing out the caveat in the description.

- `make -j6 Galacticus.exe` — **succeeded**, zero errors, `Galacticus.exe` linked.
- `./Galacticus.exe parameters/quickTest.xml` — **exit 0**, no failure markers (`FAIL`/`FAILED`/`fatal`/`aborted`/`ODE integration failed`/`unrecognized parameter`), `galacticus.hdf5` written.

The run emits one warning — "The Johnson et al. (2021) scale radius energy model was constructed but never applied to any node…" — which is pre-existing: that text is on `master` verbatim, and this branch has no diff in that file.

That completes the `CONTRIBUTING.md` pre-submission checklist.

PR #1351, opened 2026-08-10
Fix the impossible likelihood in the COZMIC WDM 3 keV X8 subhalo mass function

abensonca:
`darkMatterOnlySubhalosCOZMICWDM3keVMilkyWayX8` / `subhaloMassFunction` has returned `logImpossible` on **every run since 2026-07-29** (7 of 82 runs overall), and the dashboard did not show it, for two independent reasons. This fixes both.

### The model: 8 trees is not enough

The parameter file set `replicationCount = 4`, and with 2 host halos that builds only **8 trees** — the lowest of the twelve COZMIC WDM models, for the *warmest* dark matter particle, which produces the *fewest* subhalos.

The failure is the same one fixed for the 4 keV X8 model in #1342: the `massRatio = 0.136` bin, where the target holds a single LMC-analog subhalo, comes out empty, and an empty bin with a non-zero target makes the likelihood impossible (`source/output/analyses/subhalo_mass_function.F90:647`).

Sizing the rate from the `resolutionX1` model of the *same* 3 keV scenario, which builds 180 trees and finds 15 subhalos in that bin:

| | trees | rate/tree | expected count | P(empty) |
|---|---|---|---|---|
| now | 8 | 0.083 | 0.67 | ~51% |
| `replicationCount = 32` | 64 | 0.083 | 5.3 | 0.7% |
| **`replicationCount = 64`** | **128** | **0.083** | **10.7** | **0.011%** |

The 0.083/tree rate is consistent with the 4 keV X8 measurement of 0.094/tree. At the low end of its uncertainty (0.06/tree) 128 trees still give P(empty) ≈ 0.05%. The change that landed on 2026-07-29 reduced the abundance of massive subhalos across the COZMIC models — visible as a step in several likelihoods on that date — and pushed this model over the edge.

I checked the other models for the same exposure. Every X8 model has a target of 1 in that bin with a model rate of 0.04–0.13 per tree, but only 3 keV X8 has ever actually failed (Symphony X64 failed 6 times and already carries `ignoreEmptyModelBins`), so no other tree counts are changed here.

**CI cost:** this is by a wide margin the cheapest of the twelve COZMIC WDM jobs, 7.7 min against 202.9 min for the 10 keV X8 job — which already builds 128 trees. Even assuming all 7.7 min scales with the tree count, the job stays well inside both this job family's span and the workflow critical path (287.9 min, the decaying dark matter job), so **CI wall-clock is unchanged**.

### The dashboard: a threshold bootstrapped from the sentinel

`bootstrap_thresholds` scaled `current` by the warn/fail fractions with no guard on the value. When `current` is `logImpossible` (`-1.0d-06 × huge(1.0d0)`, about `-1.8e302`), applying a fraction to its *magnitude* puts both thresholds **below** it:

```yaml
      current: -1.7976931348623154e+302
      threshold_fail: -2.1572317618347786e+302
      threshold_warn: -1.977462448348547e+302
```

The entry could then never fail — not on any finite likelihood however bad, and not on a repeat of the impossible value either. That is why the dashboard has been reporting this analysis as `ok` throughout. It was the only entry in that state, and the 2026-08-01 bootstrap created it simply because it happened to run against a published run in which this analysis was impossible.

`bootstrap_thresholds` now skips such analyses and names them on stderr, so they are set by hand once a finite likelihood exists.

### Verification

* `--bootstrap-thresholds` against a `gh-pages` checkout writes 111 analyses instead of 112, with `warning: not bootstrapping 1 analyses whose published logL is impossible; ... darkMatterOnlySubhalosCOZMICWDM3keVMilkyWayX8/subhaloMassFunction`.
* The ordinary build path is unaffected.
* `python3 -m pytest`: 894 passed, 1 skipped.
* The parameter file parses and yields `replicationCount = 64`.

### Companion change

The threshold entry itself is re-anchored in #1350, on its 15 finite published values. That makes the metric report **fail** rather than `ok` until this PR lands and a run publishes a finite likelihood — which is the correct reading of the current state.

### DIFF
--- a/scripts/build/buildGhPagesIndex.py
+++ b/scripts/build/buildGhPagesIndex.py
@@ -23,6 +23,7 @@ Run with::
 
 import argparse
 import json
+import math
 import re
 import sys
 from datetime import datetime, timezone
@@ -31,6 +32,13 @@ from pathlib import Path
 import yaml
 
 REPO_URL = "https://github.com/galacticusorg/galacticus"
+
+# Magnitude at or above which a published logL is taken to be Galacticus'
+# ``logImpossible`` sentinel rather than a real likelihood. The sentinel is
+# ``-1.0d-06 * huge(1.0d0)`` (about -1.8e302; see
+# source/models/likelihoods/constants.F90), so any threshold comfortably
+# between the largest plausible likelihood and that value will do.
+LOG_LIKELIHOOD_IMPOSSIBLE = 1.0e100
 
 JS_PREFIX_RE = re.compile(
     r"^\s*window\.[A-Z_]+\s*=\s*", flags=re.MULTILINE
@@ -300,8 +308,15 @@ def bootstrap_thresholds(gh_root, manifest, fraction_warn=0.10, fraction_fail=0.
     """Read each ``standardLogLikelihood`` metric's currently-published
     logL values and derive per-analysis warn/fail thresholds at
     ``current - fraction * |current|``. Returns a nested dict keyed by
-    metric suffix, then analysis name."""
+    metric suffix, then analysis name.
+
+    Analyses whose published logL is Galacticus' ``logImpossible`` sentinel
+    (``-1.0d-06 * huge(1.0d0)``, i.e. about -1.8e302) are skipped rather than
+    bootstrapped. Scaling that sentinel by a fraction would place both
+    thresholds below it, leaving an entry that no finite likelihood - however
+    bad - could ever fail, and so silently disabling the analysis."""
     out = {}
+    skipped = []
     for suffix, m in (manifest.get("metrics") or {}).items():
         if m.get("aggregator") != "standardLogLikelihood":
             continue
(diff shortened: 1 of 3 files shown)


abensonca:
Added 9da49a81a, which fixes the SIGFPE that was failing the `Validate-Dark-Matter-Only-Subhalos-COZMICWDM (WDM3keV, X8)` job on this branch.

The FPE reported at `cosmology/functions/matter_lambda.p.F90` was a symptom. The prompt cusp collapse-time root finder had multiplicative upward range expansion with no upward limit; `sigma_0(t)` saturates once the Universe is Lambda-dominated, so halos whose `sigma_0,collapse` exceeds that asymptote have no root and the search doubled the trial time without bound until `a**3` overflowed in `matterLambdaHubbleParameterEpochal`. Bounding the expansion at ten times the present age converts the runaway into a clean out-of-range status; the growth factor is within ~1e-9 of its asymptote there, so no collapse time that is currently found changes.

This is a latent bug that the tree count increase exposed rather than caused. It only reaches halos at the mass resolution at late times - in this model, nodes of exactly 5.03e6 M_solar at t = 6.6-12.9 Gyr - and 3 keV X8 is the warmest particle at the finest resolution, hence the only configuration close enough to the condition. With 8 trees the tail was never sampled; with 128 it is.

Verified locally by running this branch to completion: 128/128 trees, no floating point exception, `subhaloMassFunction` logL = -35.06, i.e. finite rather than `logImpossible`. Note that run used a full build rather than the DMO build CI uses, so the absolute likelihoods are not comparable to the CI values - the meaningful results are that it completes and that the likelihood is finite.

Two pre-existing defects found along the way, both filed separately rather than fixed here: #1359 (`Warn()` appends in O(N) under a global critical section) and #1360 (`rootFinderFind` out-of-range exits leave the GSL abort handler disabled - worth noting because the fix here is the first caller to take that path routinely).

PR #1361, opened 2026-08-11
Update the solitonNFWHeated FDM model with tidal stripping

yzhaoastro:
## Summary
- Update the DMO implementation of the `solitonNFWHeated` model.
- Add the Du et al. (2018) tidal stripping model for FDM.
- Add a limit to the NFW tidal stripping treatment in solitonNFWHeated, requiring the tidal radius to remain larger than the soliton radius so that tidal stripping does not directly affect the soliton core.
- Fix the concentration model configuration in the parameter file.
- Add a validation test for the `solitonNFWHeated` FDM model.

## Type of change
- [x] Bug fix
- [x] New feature
- [ ] Docs / tooling
- [ ] Other:

## How to test
- Run the `solitonNFWHeated` FDM validation test and verify that the model completes successfully.
- Confirm that the updated DMO `solitonNFWHeated` model produces the expected validation results.

## Checklist
- [x] I ran relevant tests (or explain why not): `solitonNFWHeated` FDM validation test.
- [x] If this PR is from a fork: I enabled 'Allow edits from maintainers'


abensonca:
A few notes on things I changed while reviewing this:

1. The tidal radius limiter was checking the type of the object returned by `node%massDistribution()` and acting if that was `massDistributionSolitonNFWHeated`. But that method always returns a `massDistributionComposite,` so it wasn't triggering correctly. I now just have it read `solitonRadiusSoliton` directly (non-zero only if we have a soliton).
2. As discussed, the transition to soliton-only is now triggered by `r_tidal <= r_soliton`.
3. Some changes to the reference parameter files:
   i. The `satelliteTidalMassLoss` operator remained in the heated parameter file alongside the soliton operator, so it was applying mass loss twice.
   ii. I added comments to the parameter files to explain why they do what they do.
4. Instead of triggering destruction at `massCore <= 0`, I added a parameter that specifies the fraction of the initial core mass at which to destroy (defaulting to 1%). I think this will be more reliable since we could only get `massCore <= 0` by ODE overshoot - while reaching some fixed fraction can happen physically.
5. The core mass loss rate now accounts for the `randomOffset` (otherwise the fractional rate of mass loss would be too low for a model with `randomOffset > 0`, etc.)
6. I changed the core-halo evolution so that it does not apply to subhalos - they were getting the cosmological evolution part of this operator. I think the evolution should only apply to things before they become subhalos, but if you think it should still apply to subhalos too, we can discuss the best approach.

I also changed a few file and parameter names to make them more informative. For parameter names, these are:
| Old | New |
|---|---|
| `satelliteConditionalMassLoss` | `satelliteTidalMassLossSoliton` |
| `satelliteTidalStrippingRadius` value `limited` | `solitonLimited` |
| `conditional.F90` | `tidal_soliton.F90` |
| `Limited.F90` | `soliton_limited.F90` |

The `parametersMigrate.py` script should make these changes as needed.

Some changes to the soliton mass loss operator:

- New parameter `fractionMassCoreDestruction` (default `0.01`) - the fraction of the core mass at the soliton-only transition below which the satellite is destroyed.
- It now requires a `satelliteTidalStrippingRadius`, used to detect the transition.
- Its outer stripping model is no longer given inline: it is inherited from the enclosing
  `satelliteTidalStripping`, so it automatically picks up the limiter and any future change to the model's stripping.

I made the set of states for `solitonStatus` be an enumeration, so instead of setting values directly we can use labels - this makes it more obvious what is being set/tested in each case. _But_, it does mean that the NFW-only case is now `3` instead of `-1`. 

And, for a halo in which no soliton formed, `solitonRadiusCore` and `solitonDensityCore` are output as zero, while `solitonRadiusSoliton` and `solitonMassCore` are negative.

yzhaoastro:
Thanks for going through this so carefully and for explaining all the changes. They all make sense to me. I especially agree with stopping the core–halo evolution once a halo becomes a satellite, since its core evolution should then be driven by tidal stripping rather than cosmological evolution.
The new destruction criterion also makes sense, especially since it avoids relying on numerical overshoot in the ODE integration. The changes to the soliton-only transition and tidal radius limiter make sense to me as well. Thanks also for catching the double mass-loss issue and cleaning this up!

abensonca (approved):

Issue #1332, opened 2026-08-06
Compute the specific angular momentum of accreted material self-consistently in physical halo angular momentum models


abensonca:
## Summary

Branch `fixAngularMomentumAccretion` (commit `ed63298ee`) introduced a
`angularMomentumSpecificAccreted` property on the `spin` component, set by
`nodeOperatorHaloAngularMomentumInterpolate`, and used by
`nodeOperatorCGMAccretion` to compute the rate of angular momentum accretion
into the CGM. That change fixes a real bug (angular momentum accretion was
being silently discarded whenever `basic%accretionRate() == 0`), but the two
estimators it offers are both *reconstructions* of a specific angular momentum
from halo-level bookkeeping:

| `angularMomentumSpecificGrowthFromAccretionRate` | estimator | comment |
|---|---|---|
| `.true.` (default) | `j = J̇ / Ṁ_unresolved` | numerator includes resolved-merger contributions, denominator does not — inconsistent, and blows up as `Ṁ_unresolved → 0` |
| `.false.` | `j = ΔJ / ΔM` (child → parent) | numerator and denominator consistent, but attributes merger-borne angular momentum to smoothly accreted gas |

Both are appropriate only where the halo angular momentum is *not* determined
by the structure of the merger tree — i.e. for
`nodeOperatorHaloAngularMomentumRandom` and
`nodeOperatorHaloAngularMomentumRandomWalk`, where the notion of "angular
momentum arriving with unresolved accretion" has no meaning and the estimator
is only a plausibility device. For a physical model such as
`nodeOperatorHaloAngularMomentumVitvitska2002`, the angular momentum carried
by unresolved accretion **is** an explicit, separable term in the model, and we
should use it directly rather than reverse-engineering it from `ΔJ/ΔM`.

## Where the code stands today

- `source/objects/nodes/components/dark_matter_halo/spin/scalar.F90` — new
  `angularMomentumSpecificAccreted` property (settable/gettable, non-evolvable);
  inherited by the `vector` implementation.
- `source/nodes/operators/physics/halo_angular_momentum_interpolate.F90:135-208` —
  the only place the property is set, in `nodeInitialize`, plus propagation on
  promotion at `haloAngMomInterpolateNodePromote:295`.
- `source/nodes/operators/physics/circumgalactic_medium/accretion.F90:556-570` —
  the only consumer: `Ĵ_hot = j × Ṁ_hot`, `Ĵ_cold = j × Ṁ_cold`.
[...]

PR #1364, opened 2026-08-11
fix(nodes): use specific angular momentum for CGM accretion rates

abensonca:
## Summary

The CGM accretion operator previously reconstructed the specific angular momentum of accreted material as the ratio of the halo angular momentum growth rate to the halo mass accretion rate, and skipped the calculation entirely whenever that accretion rate was zero. Since the angular momentum growth rate is computed from the full child-to-parent change (including contributions from resolved mergers) while the accretion rate covers only unresolved accretion, that ratio is inconsistent — and the zero-accretion-rate guard silently discarded angular momentum accretion, which matters most in exactly the regime where the halo's unresolved mass growth is tiny.

The specific angular momentum of accreted material is now computed once, where the interpolation is performed, and stored in a new `angularMomentumSpecificAccreted` property of the scalar spin component (inherited by the `vector` implementation). `nodeOperatorCGMAccretion` simply multiplies it by the hot- and cold-mode mass accretion rates.

## Changes

- **`spin/scalar.F90`** — new `angularMomentumSpecificAccreted` property (settable/gettable, non-evolvable).
- **`halo_angular_momentum_interpolate.F90`** — computes and sets the property in `nodeInitialize`, propagates it on promotion, and gains a new `angularMomentumSpecificGrowthFromAccretionRate` parameter selecting between the two estimators.
- **`circumgalactic_medium/accretion.F90`** — consumes the property instead of reconstructing the ratio; the `basic%accretionRate() /= 0` guard is gone.

### The new parameter

| `angularMomentumSpecificGrowthFromAccretionRate` | estimator |
|---|---|
| `.true.` (default) | `j = J̇ / Ṁ`, from the dark-matter-only accretion rate, which excludes resolved mergers |
| `.false.` | `j = ΔJ / ΔM`, from the net child-to-parent mass growth, which includes mergers and so matches how `J̇` itself is derived |

The default is `.true.` to preserve the previous behaviour. `.false.` is arguably the more physical of the two, and is the setting that behaves sensibly when the halo's unresolved mass growth rate is small.

## Scope and follow-up

Both estimators reconstruct a specific angular momentum from halo-level bookkeeping, which is the right thing to do for angular momentum models where the halo angular momentum is *not* determined by the structure of the merger tree — `nodeOperatorHaloAngularMomentumRandom` and `nodeOperatorHaloAngularMomentumRandomWalk`. For a physical model such as `nodeOperatorHaloAngularMomentumVitvitska2002`, the angular momentum arriving with unresolved accretion is an explicit, separable term in the model and should be used directly rather than reverse-engineered.

That redesign is deliberately **not** part of this PR; it is written up in detail in #1332, including the observation that the Vitvitska operator already computes the quantity we want (`⟨j_orb⟩ × f_subres`, independent of the unresolved mass) and then discards it.

## Verification

- Full build clean (`make -j2 Galacticus.exe`, exit 0, no compiler errors).
- `parameters/quickTest.xml` smoke test runs to completion, exit 0, no failure markers in the log.
- `testSuite/test-molecular-cooling.py` passes, with a maximum deviation from its reference values of 0.5% (H₂ mass) and 0.7% (cooling function), against a 1% tolerance.
- Test models confirm the fix behaves well.

## Subsequent commits

`af80bb5e7` renames the property from `angularMomentumGrowthSpecific` (which parses as "the specific growth of angular momentum") to `angularMomentumSpecificAccreted`. It is a pure mechanical rename of a property that only exists on this branch, plus a one-word typo fix in the operator description. The name is 31 characters, matching `angularMomentumVectorGrowthRate`, so the continuation-line alignment in the interpolate operator is unchanged.

`9b87ebb69` fixes two problems in the `angularMomentumSpecificGrowthFromAccretionRate = .true.` estimator, both found by `test-molecular-cooling.py`, whose molecular hydrogen masses and cooling functions had shifted by 5–13%:

1. **The accretion-rate guard is restored to `/= 0`.** Guarding on `> 0` meant a halo losing mass removed gas from the CGM without removing that gas' angular momentum, raising the CGM's specific angular momentum unphysically.
2. **The specific angular momentum is computed from the net change in `|J|` across the branch segment, not from the `angularMomentumGrowthRate` property.** For a vector spin component that property is a deferred function returning the instantaneous projection `d|J|/dt = (J·dJ/dt)/|J|`, which varies along the segment as the angular momentum vector reorients. Sampling it once at node initialization froze it at the segment start; since `|J(t)|` is convex, that value always lies below the segment mean, so the CGM was systematically under-supplied with angular momentum. The net change in magnitude integrates to exactly `Δ|J|` across the segment and is identical for scalar and vector spin components.

Issue 2 was manifest only for `componentSpin=vector`; with a scalar spin component the getter returns the stored value and the calculation was already correct, which is why it was not caught earlier.

Follow-up: #1332

### DIFF
--- a/source/nodes/operators/physics/circumgalactic_medium/accretion.F90
+++ b/source/nodes/operators/physics/circumgalactic_medium/accretion.F90
@@ -556,10 +556,10 @@ contains
          & rateChemicalsAccretionHot=   self%accretionHalo_%      accretionRateChemicals(node,accretionModeHot  )
     ! Get the rate of angular momentum accretion onto the halo.
     if (hotHalo%angularMomentumIsSettable()) then
-       spin                            =>  node%spin                         ()
-       rateAngularMomentumAccretionHot =  +spin%angularMomentumGrowthSpecific() &
+       spin                            =>  node%spin                           ()
+       rateAngularMomentumAccretionHot =  +spin%angularMomentumSpecificAccreted() &
             &                             *     rateMassAccretionHot
-       rateAngularMomentumAccretionCold=  +spin%angularMomentumGrowthSpecific() &
+       rateAngularMomentumAccretionCold=  +spin%angularMomentumSpecificAccreted() &
             &                             *     rateMassAccretionCold
        if (self%angularMomentumAlwaysGrows) then
           rateAngularMomentumAccretionHot =abs(rateAngularMomentumAccretionHot )
--- a/source/nodes/operators/physics/circumgalactic_medium/accretion.F90
+++ b/source/nodes/operators/physics/circumgalactic_medium/accretion.F90
@@ -555,14 +555,12 @@ contains
     if (self%countChemicals > 0) &
          & rateChemicalsAccretionHot=   self%accretionHalo_%      accretionRateChemicals(node,accretionModeHot  )
     ! Get the rate of angular momentum accretion onto the halo.
-    if (basic%accretionRate() /= 0.0d0 .and. hotHalo%angularMomentumIsSettable()) then
-       spin                            =>  node %spin                     ()
-       rateAngularMomentumAccretionHot =  +spin %angularMomentumGrowthRate() &
-            &                             *      rateMassAccretionHot        &
-            &                             /basic%accretionRate            ()
-       rateAngularMomentumAccretionCold=  +spin %angularMomentumGrowthRate() &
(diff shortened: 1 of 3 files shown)


abensonca:
@copilot

PR #1396, opened 2026-08-18
fix(parameters): record parameters read during node component thread initialization

### LINKED ISSUE #1377: Parameters read during node-component thread initialization are absent from the output file (and silently dropped by parametersExtract.py)
Parameters read during node-component *thread* initialization never reach the output file's `Parameters` group, because thread initialization is run against a per-thread copy of the parameter set which is deliberately created with output suppressed.

This is not limited to defaulted parameters: **values the user supplied explicitly are lost too**, and consequently `scripts/parameters/parametersExtract.py` silently produces a parameter file that is not equivalent to the one that was run.

Found while implementing the defaulted-parameter marking in #1353; the marking made the gap visible, but the gap itself long predates it.

### Mechanism

`source/tasks/evolve_forests/_class.F90:627-629` builds a per-thread parameter set and runs thread initialization against it:

```fortran
allocate(parameters)
parameters=inputParameters(self%parameters)      ! copy constructor
call Node_Components_Thread_Initialize(parameters)
```

`inputParametersConstructorCopy` (`source/utility/input_parameters.F90`) passes **`noOutput=.true.`**:

```fortran
self = inputParameters(parameters%rootNode,noOutput=.true.,noBuild=.true., &
     &                 documentManager=parameters%documentManager,document=parameters%document)
```

so the copy has no `outputParameters`. `inputParametersSubParameters` propagates that, opening an output group only `if (associated(self%outputParameters))`. Every object built beneath that copy — and every parameter it reads — is therefore invisible to the output file.

The same `parameters=inputParameters(self%parameters)` + `Node_Components_Thread_Initialize` pattern appears in at least `source/tasks/postprocess_forests.F90:272-274` and `source/tasks/halo_mass_function.F90:778-780`, so the audit should cover all call sites, not just `evolveForests`.

Suppressing output on thread copies is presumably deliberate — without it, N threads would race to write the same attributes. The problem is that nothing then records these parameters at all.

### Measured consequences

Running `testSuite/parameters/parametersExtract.xml`:

* **21 defaulted values across 10 containers** are absent from the output, including `mass`, `massType`, `componentType`, `scaleRadius`, `scaleHeight` under `componentDisk/massDistributionDisk`, and the corresponding set under `componentSpheroid/massDistributionSpheroid`.
* **Four containers are absent from the `Parameters` group entirely** — neither as groups nor as value attributes:

  | Parameter | In the parameter file | In the output |
  | --- | --- | --- |
  | `darkMatterProfile` | `value="adiabaticGnedin2004"`, with `A` and `omega` | absent |
  | `hotHaloMassDistribution` | `value="betaProfile"` | absent |
  | `mergerMassMovements` | present | absent |
  | `mergerRemnantSize` | present | absent |

  The first two are **specified explicitly by the user**, with sub-parameters, and are still lost.
* A container can be *partly* recorded, which makes the gap easy to miss: `componentDisk` and `componentSpheroid` groups do exist and do carry correctly-marked defaults (`postStepZeroNegativeMasses{defaulted}` and friends) read on the master path, while the parameters read from the thread-init copy of the *same* container are missing. The presence of a group does not imply it is complete.

All of the affected sites are builders inside `*_Thread_Initialize` routines:

| Container | Built in |
| --- | --- |
| `componentDisk`, `componentDisk/massDistributionDisk` | `Node_Component_Disk_Standard_Thread_Initialize` |
| `componentSpheroid/massDistributionSpheroid`, `mergerMassMovements`, `mergerRemnantSize` | `Node_Component_Spheroid_Standard_Thread_Initialize` (`source/objects/nodes/components/spheroid/standard/_class.F90:307-308`) |
| `darkMatterProfile` | `Node_Component_Dark_Matter_Profile_Scale_Thread_Initialize` (`source/objects/nodes/components/dark_matter_profile/scale.F90:123`) |
| `hotHaloMassDistribution` | `Node_Component_Hot_Halo_Standard_Thread_Initialize` (`source/objects/nodes/components/hot_halo/standard/_class.F90:369`) |
| `blackHoleSeeds`, `blackHoleBinaryRecoil`, `blackHoleBinaryInitialSeparation` | `Node_Component_Black_Hole_Standard_Thread_Initialize` (`source/objects/nodes/components/black_hole/standard/_class.F90:157-159`) |

### Silent data loss in `parametersExtract.py`

Because the extractor reconstructs a parameter file from the `Parameters` group, anything missing from that group is missing from the extracted file:

```
                            original XML   extracted XML
darkMatterProfile                      1               0
hotHaloMassDistribution                1               0
mergerMassMovements                    1               0
mergerRemnantSize                      1               0
```

A parameter file extracted from an output therefore **runs different physics** from the file that produced it: `darkMatterProfile` falls back to its default class rather than `adiabaticGnedin2004`, and `hotHaloMassDistribution` rather than `betaProfile`. No warning is issued.

`testSuite/test-parameters-extract.py` does not catch this because it compares two *extracted* files, which are equally lossy and therefore identical. A check comparing an extracted file against the original would catch it, and is worth adding once the underlying gap is closed.

### Proposed solution

Give exactly one thread's copy an output group, rather than none.

Add an optional `noOutput` argument to `inputParametersConstructorCopy`, defaulting to `.true.` so every existing caller is unaffected, and have the thread-initialization call sites pass `noOutput=.false.` on the master thread only:

```fortran
!$omp critical(evolveForestsInitialize)
allocate(parameters)
!$ if (OMP_Get_Thread_Num() == 0) then
!$    parameters=inputParameters(self%parameters,noOutput=.false.)
!$ else
      parameters=inputParameters(self%parameters)
!$ end if
call Node_Components_Thread_Initialize(parameters)
!$omp end critical(evolveForestsInitialize)
```

This is attractive because:

* **No new locking.** The call site is already inside `!$omp critical(evolveForestsInitialize)`, so the single writer is serialized against everything else by construction.
* **One writer, so no race and no duplicate-write error.** Every thread reads the same parameters and would write identical values, so recording from one is sufficient. Attribute writes are in any case already check-then-write under `hdf5Access`, hence idempotent.
* **No extra object construction**, and no change to what any other thread does.

Alternatives considered:

* *Let every thread write.* Relies on the `hasAttribute`/`hdf5Access` pairing being atomic across all paths, adds locking surface, and buys nothing over a single writer.
* *Pre-populate on the master before the parallel region* by running initialization an extra time with the real parameter set. This doubles object construction and re-runs side effects such as event-hook registration; too invasive.
* *Move the builders out of node components into `nodeOperator`s*, which are built from the output-enabled parameter set. This is already planned for the black-hole classes and fixes them, but only for classes actually migrated — it does not close the general mechanism.

### Not in scope

The related wart in `inputParametersPath`, which built paths in reverse order, is fixed separately as part of #1353.

abensonca:
Closes #1377.

## Problem

Parameters read during node component *thread* initialization never reached the output file's `Parameters` group. Thread initialization runs against a per-thread copy of the parameter set built by `inputParametersConstructorCopy`, which passed `noOutput=.true.` unconditionally, so the copy had no output group and every object built beneath it — and every parameter it read — was invisible to the output file.

This was not limited to defaulted values. Parameters supplied explicitly by the user were lost too, so `scripts/parameters/parametersExtract.py` silently produced a parameter file that was *not* equivalent to the one that was run: `darkMatterProfile` fell back to its default class rather than `adiabaticGnedin2004`, and `hotHaloMassDistribution` rather than `betaProfile`. No warning was issued.

## Fix

Give exactly one thread's copy an output group, rather than none.

`inputParametersConstructorCopy` gains an optional `noOutput` argument, defaulting to `.true.` so every existing caller is unaffected. Passing `noOutput=.false.` shares the output group of the parameter set being copied; the resource managers are assigned rather than pointer-assigned, so the reference counts of the shared HDF5 objects are incremented and the group stays valid for the copy's lifetime.

`taskEvolveForests` and `taskPostprocessForests` pass `noOutput=.false.` on the master thread only. Every thread reads the same parameters, so a single writer suffices. No new locking is required: in `taskEvolveForests` the call site is already inside `critical(evolveForestsInitialize)`, and elsewhere the single writer's output passes through the existing HDF5 access lock, under which attribute writes are already check-then-write and hence idempotent.

The condition is written as

```fortran
recordParameters=.true.
!$ recordParameters=OMP_Get_Thread_Num() == 0
parameters=inputParameters(self%parameters,noOutput=.not.recordParameters)
```

rather than an `!$`-guarded `if`/`else`, so that a build without OpenMP still records the parameters.

## Audit of the remaining call sites

Both are left suppressed, with comments recording why:

| Call site | Why output stays suppressed |
| --- | --- |
| `taskHaloMassFunction` | Already performs thread initialization on the output-enabled `self%parameters` earlier in `perform`, so these parameters are recorded there. Its per-thread copy is also never deallocated, so enabling output would leak a reference. |
| `mergerTreeEvolverThreaded` | Its parallel region is nested inside that of the driving task, so the team's master thread is not unique — enabling output would give one writer per outer thread, repeated for every tree evolved. It also uses the same root parameter set the driving task already records. |

## Test

`testSuite/test-parameters-extract.py` gains a check comparing the extracted parameter file against the **original**. The existing check compares two *extracted* files, which are equally lossy and therefore agree even when parameters are missing entirely.

The comparison is by (path, value) multiset, so a parameter must appear at the same point in the hierarchy and repeated instances must be repeated in both files. Values are normalized numerically, since a value written verbatim to the output may be reformatted when read back. The deliberately-duplicated `<mergerTreeOutputter value="null"/>` is the single documented exclusion.

To confirm the check is a genuine regression test rather than a tautology, I ran it against the `--dry-run` extraction — which never calls `task_%perform` and so reproduces the pre-fix state. It reports 17 missing user-supplied parameters, including `darkMatterProfile = adiabaticGnedin2004`, `hotHaloMassDistribution = betaProfile`, `mergerMassMovements = simple`, and `mergerRemnantSize = cole2000`.

## Verification

Built clean. `testSuite/test-parameters-extract.py`, `testSuite/test-postprocessing.py`, and `parameters/quickTest.xml` all pass.

In the `quickTest` output, all four previously-absent containers are now present, and `componentDisk/massDistributionDisk` carries its five previously-missing defaulted values with correct `{defaulted}` markers:

```
['componentType', 'componentType{defaulted}', 'dimensionless', 'mass', 'massType',
 'massType{defaulted}', 'mass{defaulted}', 'scaleHeight', 'scaleHeight{defaulted}',
 'scaleRadius', 'scaleRadius{defaulted}']
```

---

This change was generated with Claude and verified by Andrew Benson.

### DIFF
--- a/source/merger_trees/evolver/threaded.F90
+++ b/source/merger_trees/evolver/threaded.F90
@@ -421,6 +421,12 @@ contains
        timer_=timer()
        call timer_%start()
     end if
+    ! The per-thread copies of the parameter set are made with output suppressed (the default), so parameters read during thread
+    ! initialization are not recorded in the output file from here. That is deliberate: this parallel region is nested inside the
+    ! parallel region of the task driving the evolution, so the master thread of this team is not unique---enabling output on it
+    ! would give one writer per outer thread, and would repeat that write for every tree evolved. It is also unnecessary, since
+    ! `self%parameters` is the root parameter set, on which the driving task has already performed thread initialization with
+    ! output enabled, so exactly these parameters are recorded there.
     allocate(parameters)
     parameters=inputParameters(self%parameters)
     call Node_Components_Thread_Initialize(parameters)   
--- a/source/tasks/evolve_forests/_class.F90
+++ b/source/tasks/evolve_forests/_class.F90
@@ -623,17 +623,31 @@ contains
     ! Perform initialization which must occur for all threads if run in parallel. This is done first---before the per-thread deep
     ! copies below---so that the tree census (which, for some mass distributions, requires building nodes) can be computed on the
     ! master constructor and then inherited by each thread's copy.
-    !$omp critical(evolveForestsInitialize)
-    allocate(parameters)
     ! Thread initialization is run against a per-thread copy of the parameter set. Output is enabled on the master thread's copy
     ! only, so that the parameters read during thread initialization (some of which---e.g. the mass distributions of the disk and
     ! spheroid components---are read nowhere else) are recorded in the output file. Every thread reads the same parameters, so a
-    ! single writer suffices, and the enclosing critical section serializes that writer against everything else.
+    ! single writer suffices.
+    !
+    ! The master must initialize *first*, and alone. Building an object from its default class inserts that default into the
+    ! parameter tree---which every thread's copy shares---so only the thread which initializes first sees such a parameter as
+    ! absent, and only that thread marks it as defaulted. Were the master not that thread, those markers would never be written,
+    ! and which parameters carried one would depend on the number of threads.
     recordParameters=.true.
     !$ recordParameters=OMP_Get_Thread_Num() == 0
-    parameters=inputParameters(self%parameters,noOutput=.not.recordParameters)
-    call Node_Components_Thread_Initialize(parameters)
-    !$omp end critical(evolveForestsInitialize)
+    if (recordParameters) then
+       allocate(parameters)
+       parameters=inputParameters(self%parameters,noOutput=.false.)
+       call Node_Components_Thread_Initialize(parameters)
+    end if
+    !$omp barrier
+    ! The remaining threads now initialize against copies with output suppressed.
+    if (.not.recordParameters) then
+       !$omp critical(evolveForestsInitialize)
+       allocate(parameters)
+       parameters=inputParameters(self%parameters)
+       call Node_Components_Thread_Initialize(parameters)
+       !$omp end critical(evolveForestsInitialize)
+    end if
     !$omp barrier
     ! Compute the tree census and (if a cost model is available) the total predicted work, then report the start-of-run estimate.
     ! This is performed on the master constructor, before the per-thread deep copies below, so that each thread's copy inherits the
--- a/source/tasks/evolve_forests/_class.F90
+++ b/source/tasks/evolve_forests/_class.F90
@@ -520,14 +520,14 @@ contains
     !$omp threadprivate(event_)
     type            (treeNode                ), pointer                  , save :: node
     class           (nodeComponentBasic      ), pointer                  , save :: basic
-    logical                                                              , save :: treesFinished
+    logical                                                              , save :: treesFinished        , recordParameters
     integer         (c_size_t                )                           , save :: treeNumber
     type            (inputParameters         ), allocatable              , save :: parameters
     integer         (c_size_t                )                                  :: treeCount
     integer         (omp_lock_kind           )                                  :: initializationLock
     integer         (kind_int8               )                                  :: systemClockRate      , systemClockMaximum
     integer                                                              , save :: statusForest
-    !$omp threadprivate(node,basic,treeNumber,treesFinished,parameters,statusForest)
+    !$omp threadprivate(node,basic,treeNumber,treesFinished,parameters,statusForest,recordParameters)
     logical                                                                     :: checkpointRestored   , checkpointing         , &
          &                                                                         universeUpdated      , failed
     ! Whole-run progress and run-time estimation accumulators. The counts, work sums, and clock references below are shared across
@@ -625,7 +625,13 @@ contains
     ! master constructor and then inherited by each thread's copy.
     !$omp critical(evolveForestsInitialize)
     allocate(parameters)
-    parameters=inputParameters(self%parameters)
+    ! Thread initialization is run against a per-thread copy of the parameter set. Output is enabled on the master thread's copy
+    ! only, so that the parameters read during thread initialization (some of which---e.g. the mass distributions of the disk and
+    ! spheroid components---are read nowhere else) are recorded in the output file. Every thread reads the same parameters, so a
+    ! single writer suffices, and the enclosing critical section serializes that writer against everything else.
+    recordParameters=.true.
+    !$ recordParameters=OMP_Get_Thread_Num() == 0
+    parameters=inputParameters(self%parameters,noOutput=.not.recordParameters)
     call Node_Components_Thread_Initialize(parameters)
     !$omp end critical(evolveForestsInitialize)
     !$omp barrier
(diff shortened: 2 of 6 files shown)


abensonca:
CI caught a real defect in the first commit, now fixed in 2d600747c.

**What went wrong.** Giving the master thread's copy an output group is necessary but not sufficient — the master must also be the *first* thread to use its copy.

Building an object from its default class calls `parameters%addParameter` to insert that default into the parameter tree (`Process/FunctionClass/__init__.py`), and every thread's copy shares that tree (`self%parameters => parameters%parameters`). Only the thread which initializes first therefore sees such a parameter as absent and calls `markDefaulted`; a thread arriving later finds it present, writes its value, and never marks it.

With initialization order unconstrained, the `{defaulted}` markers for the classes built from their defaults during thread initialization were written only when the master happened to win the race. `Miscellaneous (test-parameters-default-reporting.py)` caught it via its thread-count invariant:

```
FAILED: the markers differ between 1 and 4 threads (4 differences).
First few: [('parameters', 'blackHoleBinaryInitialSeparation'),
           ('parameters', 'blackHoleBinaryRecoil'),
           ('parameters', 'blackHoleSeeds'),
           ('parameters', 'mergerProgenitorProperties')]
```

All four appear at 1 thread and not at 4, with no difference in the value attributes — which is why only the marker comparison caught it.

**Fix.** The master now initializes first and alone, then a barrier, then the remaining threads. The recording copy is thereby also the first copy used. The requirement is documented on `inputParametersConstructorCopy`, since it is now part of the contract for `noOutput=.false.`.

**Verification.** Locally, `test-parameters-default-reporting.py` now reports 211 markers (up from 207 — the four that were being lost), identical on 1 and 4 threads; I repeated the 1-vs-4 comparison three times to confirm it is stable rather than lucky. `test-parameters-extract.py`, `test-postprocessing.py`, and `parameters/quickTest.xml` all still pass.

Worth noting the pre-existing behavior this exposes: because materializing a default mutates the shared parameter tree, the `{defaulted}` marking has always been order-dependent — it was simply invisible while no thread-initialization copy had an output group at all.

PR #1398, opened 2026-08-18
fix(build): re-preprocess when the preprocessor's own Python sources change

### LINKED ISSUE #1395: Preprocess rules omit the Python generator sources, so editing a code generator silently leaves generated Fortran stale
## Summary

The rules that produce preprocessed Fortran do not list the Python code that
produces it among their prerequisites. Editing a source-tree processor,
a `SourceTree/Process/*` module, or `preprocess.py` itself therefore does
**not** cause any `.p.F90` / `.p.Inc` to be regenerated on an incremental
build. `make` reports success, the binary relinks, and the generated Fortran
is silently stale.

This makes "change a generator, rebuild, compare the generated code" — the
natural way to verify a code-generation change — quietly produce a false
negative.

## The rules

`Makefile:293`:

```make
$(BUILDPATH)/%.p.F90.up : source/%.F90 $(BUILDPATH)/hdf5FCInterop.dat $(BUILDPATH)/openMPCriticalSections.xml $(BUILDPATH)/stateStorables.xml $(BUILDPATH)/deepCopyActions.xml $(BUILDPATH)/directiveLocations.xml
	./scripts/build/preprocess.py source/$*.F90 $(BUILDPATH)/$*.p.F90
```

The prerequisites cover the Fortran input and the directive catalogs, but not
`./scripts/build/preprocess.py`, nor anything under `python/Galacticus/Build/`
that it imports. The same omission is in the sibling rules:

| Location | Rule |
| --- | --- |
| `Makefile:293` | `$(BUILDPATH)/%.p.F90.up` |
| `Makefile:539` | `$(BUILDPATH)/%.p.Inc.up` |
| `Makefile:711` | `$(BUILDPATH)/libgalacticus.p.Inc.up` |

The convention is already established elsewhere in the same file —
`Makefile:709` lists its implementation modules explicitly:

```make
$(BUILDPATH)/libgalacticus.Inc: ... ./scripts/build/libraryInterfaces.py ./python/LibraryInterfaces/Pipeline.py ./python/LibraryInterfaces/Emitters.py ./python/LibraryInterfaces/ArgSpec.py ./python/LibraryInterfaces/Hierarchy.py ./python/LibraryInterfaces/Classification.py
```

so the preprocess rules are the outlier, not the norm.

## Reproducer

Observed while verifying #1381, which changes the derived-type opener regex
used by the source-tree parser — i.e. code that runs inside `preprocess.py`:

1. Full clean build from a wiped `$(BUILDPATH)`.
2. Edit nine Python files, including `python/Fortran/Utils.py` and three
   modules under `python/Galacticus/Build/SourceTree/`.
3. `make -j48 Galacticus.exe`

Result: the build log is **6 lines**, with **0** invocations of
`preprocess.py` and 2 `gcc` compiles (the build-metadata objects only). Every
`.p.F90` on disk is still the one generated in step 1.

Diffing the generated tree against the pre-change snapshot then reports
"1791 files, 0 differences" — not because the change is inert, but because
nothing was regenerated. Only wiping `$(BUILDPATH)` and rebuilding produces a
real answer (which in that case was, genuinely, no change — but that was not
knowable from the incremental build).

## Why this matters

- It is silent. There is no warning, and the failure mode looks exactly like
  a correct "no change" result, so it actively misleads.
- It affects the verification step for *any* code-generation change, which is
  precisely the class of change that most needs verifying.
- A stale `.p.F90` can also be compiled into a shipped binary: if a generator
  fix is applied and the tree is not wiped, the fix is simply absent from the
  build while `make` reports success.

## Suggested approach

Add the generator implementation to the prerequisites of the three rules, most
maintainably via a variable, e.g.

```make
PREPROCESSSOURCES := ./scripts/build/preprocess.py $(shell find ./python -name '*.py')
```

Two things to weigh:

- **Granularity.** The whole of `python/` is a blunt dependency — touching any
  Python file would re-preprocess the entire tree. That is safe but slow.
  Narrowing to the modules `preprocess.py` actually imports is faster but has
  to be kept accurate, and getting it wrong reintroduces the silent staleness.
  Given that the `.up` sentinel already limits *recompilation* to files whose
  preprocessed output really differs, the blunt version's cost is mostly the
  re-preprocess sweep rather than a full rebuild.
- **`__pycache__` / editor backups** must be excluded from any `find`, or the
  prerequisite list will churn.

## Notes

Found while verifying #1381; that fix is unaffected (it was re-verified with a
forced clean rebuild) and this is independent of it.

abensonca:
Closes #1395.

## The problem

The rules producing preprocessed Fortran listed the Fortran input and the directive catalogs among their prerequisites, but nothing about the Python program that does the preprocessing. Editing `preprocess.py`, a `SourceTree/Process` hook, or any other module in its import closure therefore regenerated no `.p.F90` / `.p.Inc` at all: `make` reported success, the binary relinked, and the stale generated Fortran was compiled. It also made "change a generator, rebuild, and diff the generated code" — the natural way to verify a code-generation change — report a false "no change".

## The fix

`scripts/build/preprocessorSources.py` (new) walks the import graph of `scripts/build/preprocess.py` statically with `ast` — nothing is imported, so no side effects are triggered — resolving module names against `python/` and the entry script's own directory, and writes `$(BUILDPATH)/preprocessorSources.digest`, one MD5 per file in the closure, only-if-changed via `FileChanges.update(..., prove_update=True)`. The three preprocessing rules (`%.p.F90.up`, `%.p.Inc.up`, `libgalacticus.p.Inc.up`) take that digest as a prerequisite.

The two granularities are deliberately different:

* the digest **rule** is triggered by any Python source at all (and by the directories containing them, so deleting a module is noticed), because the closure can only be known by inspecting the imports;
* the digest **content** covers only the closure, so the expensive consequence — re-running the preprocessor over every source file — follows only from editing a module the preprocessor really uses. Editing an analysis script recomputes the digest, finds it unchanged, and stops there.

`$(PYTHONSOURCES)` / `$(PYTHONDIRS)` use the existing `rwildcard` / `rsubdirs` built-ins rather than `$(shell find …)`, which the `compileprof` profiler `SHELL` would corrupt (as the Makefile already documents for `$(ALLSOURCES)`), and `__pycache__` is filtered out so interpreter-bumped mtimes do not make the rule look out of date on every build.

113 of the 205 files under `python/` are in the closure. It was cross-checked against a real dynamic import of `preprocess.py`: the static closure is a strict superset of `sys.modules`, the extras being the hooks' function-level imports (e.g. the node-component generator reached from the `componentBuilder` hook), so nothing is missed.

## Verification

Full clean build from an empty `$(BUILDPATH)`, then, at `-j20`:

| Scenario | Before | After |
| --- | --- | --- |
| Edit `Process/Allocate.py`, rebuild | **0** `preprocess.py` invocations | **1785/1786** sources re-preprocessed in 28 s; no `.p.F90` content changed, so nothing recompiled |
| Generator change that alters the emitted Fortran | regenerates nothing | the changed line appears in the regenerated `.p.F90` |
| Edit `python/cloudy.py` (outside the closure) | — | digest script only, digest unchanged, nothing re-preprocessed, 3.4 s |
| No-op rebuild | 3.3 s | 3.3 s |

`./Galacticus.exe parameters/quickTest.xml` runs clean (exit 0, HDF5 written). `python -m pytest`: 962 passed, 1 skipped, including 7 new tests in `scripts/build/tests/test_preprocessor_sources.py` that pin both properties the digest must have — the closure is complete (every registered `Process` hook is in it, function-level imports included) and it is narrower than the whole of `python/`.

Not verified by a build: the `libgalacticus.p.Inc.up` rule change, since no `GALACTICUS_BUILD_OPTION=lib` build was run. It is the same one-token addition as the other two rules, and the digest rule is build-option independent.

## Follow-up (not in this PR)

The same gap exists in other generator rules — `Makefile_Directives`, `moduleDependencies.py`, `useDependencies.py` and friends list their driver script but not the `python/` modules it imports. `preprocessorSources.py` takes its entry points from a list (`ENTRY_POINTS`), so it can be pointed at them later.


abensonca:
## macOS build hang — cause and fix (68db7d0e4)

The macOS jobs on this branch hung and were killed at the 6 hour limit. Cause was in this PR, now fixed.

**What went wrong.** The digest rule listed the *directories* under `python/` alongside the files, copying the `$(SOURCEDIRS)` convention that lets the source-scanning rules notice a deleted source. That does not work for Python: the interpreter writes `__pycache__` into those directories whenever anything imports from them — including the build's own ~1800 `preprocess.py` invocations — so their mtimes advance mid-build and the rule is never satisfied. It re-arms itself.

That is more than wasted work. Because the preprocessed `_class.p.F90` is a prerequisite of the generated `Makefile_Use_Dependencies`, this digest is remade inside make's *makefile-remaking* phase, so a self-re-arming rule runs again in every restart pass instead of just the first. `/usr/bin/make` on the macOS runners is Xcode's **GNU Make 3.81**, and it does not survive that: it never reaches the main preprocessing sweep, spinning at 100% CPU in its own dependency processing (shallow stack, no syscalls in a `sample` profile — an algorithmic blowup, not a deadlock). Every macOS log stops at the identical line, right after the last `bound_functions.inc` is written.

It went unseen on Linux (make 4.x) and in working macOS checkouts, where `python/**/__pycache__` already exists so the mtimes never move. Only a pristine clone — what CI checks out — exposes it.

**The fix** is to drop the directory prerequisites. Nothing is lost: deleting a module the preprocessor imports fails loudly at the next `preprocess.py` run, and deleting one it does not import changes nothing to catch. The rule's remaining prerequisites are all files the build never writes, so its recipe now runs exactly once per clean build.

**Verified** by A/B on macOS 26.5 with Apple's GNU Make 3.81 at `-j3`, in a pristine clone with `work/build` and every `__pycache__` removed between runs: before, the build stops dead after the `.inc` batch; after, it walks straight into the main sweep. On Linux, touching every `python/**/__pycache__` and its parents now leaves make with nothing to do at all, where before it re-ran the digest script; a Process-hook edit still recomputes the digest and re-preprocesses all 1786 sources.

One caveat worth recording: an attempt to reproduce this under a locally-built GNU Make 3.81 on Linux hung on `master` too, so that reproduction was worthless as a discriminator. The A/B above, on a real Mac with Apple's own 3.81, is what actually settled it.

PR #1456, opened 2026-09-09
Link checker: scan the whole source tree, and extract URLs correctly

abensonca:
Found while working on #1455: `scripts/aux/linkChecker.py` lists `./source` with `os.listdir`, so it only ever sees files at the *top* of that directory. Since `source/` was reorganized into a directory hierarchy that is **3 files out of 2001**, and essentially every URL embedded in the source has gone unchecked — which is why the 74 dead Bugzilla links fixed in #1455 sat there unreported.

The sweep collects **1** URL today. It should collect **192**.

## The fix

Walk the tree, in a `scan_sources` function so the behavior is testable.

## What walking the tree exposed

Doing that exercised the URL *extraction* against the bulk of the source for the first time, and it turns out three forms produce a "URL" that is not one. Each would have been reported as a broken link. Of 164 URLs probed, 53 failed — all but 6 of them artifacts:

| Form | Problem | Example |
| --- | --- | --- |
| RST link target inside an XML directive | delimiters are written `&lt;`/`&gt;`, which the URL pattern cannot see, so unescaping leaves a `>` stuck on the end | `https://www.roe.ac.uk/~jap/haloes/>` |
| trailing punctuation + enclosing paren | the two interleave, and one pass of each leaves the other behind | `.../450/meta.` |
| Fortran string concatenation | the literal is only a *fragment*; no complete URL exists to check | `https://github.com/cmbant/CAMB/archive/refs/tags/` |

Concatenated fragments are skipped rather than guessed at — missing a link is much cheaper than a standing false alarm.

Two further groups were failing while working exactly as intended, so they are now excluded or tolerated:

- **RFC 2606 reserved names** (`.invalid`, `example.com`/`.net`/`.org`) — used in the source for URLs that are deliberately not real, e.g. the download system test's shell-injection payloads and its dummy download target.
- **HTTP 403 from `math.stackexchange.com`, `goldbook.iupac.org`, `journals.aps.org`** — these block the checker at the CDN (verified: still 403 with a browser-like user-agent) while serving the page to ordinary clients. This is precisely the case already handled for `openmp.org`, so that host's inline special case is folded into the same list.

With the extraction fixes and the reserved-domain exclusions in place, **6 of 139 probed URLs fail — and all 6 are real, not artifacts**:

| URL | Status |
| --- | --- |
| `icc.dur.ac.uk/.../genmf_readme.html` | 404 — dead page |
| `las.osaka-sandai.ac.jp/~inoue/ANAIGM/ANAIGM.tar.gz` | 404 — dead page |
| `drive.google.com/uc?...1DI52tMO4PEN...` | 404 — dead page |
| `math.stackexchange.com/questions/40713/...` | 403 — CDN block |
| `goldbook.iupac.org/terms/view/B00598` | 403 — CDN block |
| `journals.aps.org/prl/abstract/10.1103/PhysRevLett.121.211302` | 403 — CDN block |

The probe was run *before* the tolerated-403 list was added, which is why all 6 appear here. The three 403s are the CDN blocks described above and are now tolerated, so **3 failures remain with the final code** — the three dead pages below.

## The three dead links, now resolved

All three are fixed in the second commit, so this PR leaves no known broken link behind:

| Link | Resolution |
| --- | --- |
| `icc.dur.ac.uk/.../genmf_readme.html` | Page is gone; repointed at the [Wayback Machine capture](https://web.archive.org/web/20250711045335/http://icc.dur.ac.uk/Research/PublicDownloads/genmf_readme.html) (verified HTTP 200). |
| `las.osaka-sandai.ac.jp/~inoue/ANAIGM/ANAIGM.tar.gz` | Tarball is gone and no archived copy exists. It records the provenance of the data the test compares against, so it is kept, annotated as dead in the source, and excluded from the checker. |
| `drive.google.com/uc?...` | Removed. See below — it was dead code fetching a dead URL. |

### The Google Drive download was dead code

The test downloaded `stellarPopulations/SSP_Spectra_BC2003_lowResolution_imfSalpeter.hdf5` into the **dynamic** datasets path — but that file is present in the **static** datasets, and the test already reads it from there (`fileNameSpectra`). Nothing ever read the downloaded copy. So the download is removed, along with the now-unused `fileName` variable and the `File_Exists`/`download` imports.

One subtlety worth a reviewer's eye: `Directory_Make(fileName)` had a second effect. Given a *file* path it creates each intermediate directory — including `<dynamic>/stellarPopulations`, which is `pathStore`, where the test stores its results — and then goes on to `mkdir` the full path, creating a *directory* named `SSP_Spectra_...hdf5` (which is also why `File_Exists` then returned true and the dead download never actually fired). Removing the call therefore required creating `pathStore` explicitly, which is what was meant in the first place.

> [!IMPORTANT]
> The Fortran changes are **not compile-tested**. The development box used has gfortran 13.3.0 (Galacticus requires ≥ 16) and no HDF5 headers, so the build fails long before reaching these files. Please confirm `tests.stellar_populations.luminosities.exe` builds and passes on a suitable toolchain.

## Also: the `vX.Y.Z` placeholder

The most recent Link-Check run failed on `https://galacticus.readthedocs.io/en/vX.Y.Z/`. That appears in the "Versions and Releases" page of the developer guide, where it names the *form* of a per-release documentation URL rather than any actual release — a placeholder, not a link — so it is now excluded. A real per-version URL (`.../en/v1.2.3/`) is still checked.

## Impact on the job

In a CI checkout (where the generated documentation pages do not exist) the curl-checked URL count goes from ~294 to ~451, roughly 3 minutes more runtime. The job already uses `continue-on-error` and requires 3 *consecutive* failures before flagging a URL, so a transient failure in the newly-covered set will not turn it red.

## Verification

New unit tests (`scripts/aux/tests/test_link_checker.py`, 22 cases) cover the tree walk, file selection, the source locations recorded for each URL, all three extraction fixes, the reserved-domain exclusions and the tolerated 403s. 36 aux tests pass. Every URL the fixed sweep collects was probed with the checker's own curl options to produce the numbers above.

---

Depends on nothing, but pairs naturally with #1455 — that PR repairs the 74 dead Bugzilla URLs which this PR would otherwise start reporting.

Drafted with assistance from Claude; to be reviewed and verified by a human before merge, per [CONTRIBUTING.md](https://github.com/galacticusorg/galacticus/blob/master/CONTRIBUTING.md#ai-assisted-contributions).

https://claude.ai/code/session_01B1YXEsVCw2mGzmvucFceFm

### DIFF
--- a/scripts/aux/linkChecker.py
+++ b/scripts/aux/linkChecker.py
@@ -93,6 +93,15 @@ _EXCLUDED_URL_RES = [
     re.compile(r'^https://drive\.google\.com/open\?id=0B7vqPPPgOdtIfjUtb3RsV2JUOTFFX29WV1FZNURPMHAxTEtZQjhJOGtyNXZUTTNVSzFZazQ$'),
     # Defunct Maraston stellar population model link, retained only for the historical record.
     re.compile(r"^http://www\.icg\.port\.ac\.uk/~maraston/Claudia's_Stellar_Population_Model\.html$"),
+    # A documentation URL written with a placeholder version, naming the form of
+    # a per-release documentation URL rather than any one release's -- see the
+    # "Versions and Releases" page of the developer guide.
+    re.compile(r'^https://galacticus\.readthedocs\.io/en/vX\.Y\.Z(?:[/?#]|$)'),
+    # Defunct link to Inoue's own IGM attenuation code, from which the file that
+    # `source/tests/spectra/postprocess/Inoue2014.F90` compares against was
+    # extracted.  No archived copy of the tarball could be found, so the URL is
+    # retained only to record where that file came from.
+    re.compile(r'^http://www\.las\.osaka-sandai\.ac\.jp/~inoue/ANAIGM/ANAIGM\.tar\.gz$'),
     # RFC 2606 reserves the `.invalid` TLD, and the `example.com`/`.net`/`.org`
     # domains, for use in documentation and testing.  The source uses both for
     # URLs which are deliberately not real -- the download system test's
--- a/scripts/aux/linkChecker.py
+++ b/scripts/aux/linkChecker.py
@@ -70,6 +70,21 @@ def record_success(url, failures):
 # literals (`` ``url`` ``) and are never valid inside a URL.
 _URL_RE = re.compile(r'https?://[^\s<>"}\]`]+')
 
+# A delimiter that terminates a URL, but which may be hidden behind an entity
+# reference in the source -- an RST link target ``<url>`` embedded in an XML
+# directive writes its closing ``>`` as ``&gt;``, which the regex above cannot
+# see and which unescaping then leaves stuck to the end of the URL.
+_URL_TERMINATOR_RE = re.compile(r'[\s<>"`]')
+
+# Fortran string concatenation immediately inside, or immediately following, a
+# matched URL: ``'https://…/archive/v'//version//'.tar.gz'``.  The literal is
+# only a *fragment* of a URL which is assembled at run time, so there is no
+# complete URL here to check -- and the fragment, checked as though it were
+# one, is reported as broken.  Such URLs are skipped: missing a link is much
+# cheaper than a standing false alarm.
+_CONCATENATION_RE       = re.compile(r'[\'"]\s*//')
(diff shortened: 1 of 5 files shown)


abensonca:
`tests.stellar_populations.luminosities.exe` compiles and runs successfully.

Issue #115, opened 2020-03-14
Introduce a `table2D` base class and unify the 2D table interface


abensonca:
Multi-dimensional tables could be redesigned such that we can composite different interpolation algorithms and grid spacings in each dimension.

Then couple just use a parameterized derived type to allow arbitrary dimensionality.

abensonca:
## Assessment: rescope to a `table2D` base class; the rest is either already done or not worth doing

I went through the table classes to see how much of this proposal is still live. Splitting it into its
three constituent ideas:

### 1. Arbitrary dimensionality — already delivered, elsewhere

`Numerical_Interpolation_MultiD` (`source/numerical/interpolation/multiD.F90`) already provides
multilinear interpolation in a separable grid of arbitrary dimensionality, with **one `interpolator`
object per dimension** — so extrapolation behavior is already specified per-dimension, and a
logarithmic dimension is handled by building its interpolator on the logarithm of the coordinate.
That is the "composite different ... grid spacings in each dimension" ask, implemented at the
`interpolator` layer rather than the `table` layer. It landed as Phase 0 infrastructure for the
unified dust framework (65c3620fa) and is currently exercised only by
`source/tests/interpolation/multiD.F90`.

So the arbitrary-dimensionality half of this issue does not need doing again at the `table` layer.
If anything the follow-up is *adoption*: hand-rolled rank-4 tabulations still exist (e.g.
`source/satellites/merging/virial_orbits/loss_cone.F90`,
`source/mass_distributions/spherical/tabulated.F90`), and those are the candidates.

### 2. Parameterized derived types — recommend against

`source/` already carries **77** `<workaround type="gfortran" .../>` blocks, one of them inside
`source/objects/tables/_module.F90` itself. PDTs are historically among the least reliable corners of
gfortran's F2003/F2008 support, and `table` is a core type on which a large fraction of the code
depends. The payoff (one type instead of a small number) does not justify betting the table layer on
that. `interpolatorMultiD` reaches arbitrary dimensionality with an ordinary allocatable
`dimension(:)` component and no PDT, which is the right precedent.

### 3. Compositing scalings and interpolation algorithms — worth it in 2D, not in 1D

**1D: leave alone.** The eight concrete 1D types look like a combinatorial explosion, but every one
of them is actually used:

| type | files using |
| --- | --- |
| `table1DLogarithmicLinear` | 31 |
| `table1DGeneric` | 14 |
| `table1DLinearLinear` | 11 |
[...]

PR #1459, opened 2026-09-09
feat(tables): introduce a table2D base class and unify the 2D table interface

abensonca:
Implements steps **(a)** and **(b)** of the plan in #115, plus a fix to the memoization those steps exposed.

## The problem

`table2DLinLinLin` and `table2DLogLogLin` each extended `table` directly, so nothing could be written polymorphically over "a 2D table". Their interfaces had diverged accordingly — `size`, `x`, `y`, `z`, `isInitialized`, `interpolationFactors`, `interpolateGradient` and `extend` existed only on the latter, and `interpolatorReinitialize` only on the former — and so had their implementations, so only `table2DLogLogLin` could be extended onto an absolute lattice or cached via `Table_Caches`.

## (a) A `table2D` abstract base

Carries `xCount`, `yCount`, `xv`, `yv`, `zv`, the per-axis extrapolation types and both lattices. `interpolate` and `interpolateGradient` are deferred; the accessors which do not depend on the scaling are concrete. Both existing types are re-parented onto it and reduced to what genuinely differs — the interpolators of the one, the memoized interpolation state of the other.

## (b) Unifying the interface

- **`Table_2D_Extend_Uniform`**, the two-dimensional counterpart of `Table_1D_Extend_Uniform`, with each concrete type setting its own spacings and resetting its own cached state on return. `table2DLinLinLin` thereby gains a `perUnit` `extend` — which is what lets it participate in `Table_Caches` at all.
- **`interpolateGradient` for `table2DLinLinLin`**, including the `fix` and `zero` extrapolation cases.
- **`populate` on the base**, via a new `interpolationReset` hook.
- **`zs`**, which returned every table as a rank-3 array on one type and a rank-2 slice of one table on the other. The rank-3 form had no callers, so both now take an optional `table` and return a slice.
- **The 2D cache routines** take `class(table2D)`, dispatch on a new `tableTypeLinLinLin2D` enumeration entry, and ignore a file written by a different table type. A cached tabulation is adopted by extending onto its lattices and copying its values rather than by whole-object assignment, since the target is polymorphic and not allocatable — and its dynamic type must be preserved in any case.

## The memoization fix

A single key, `tablePrevious`, guarded *both* the memoized value and the memoized gradient of `table2DLogLogLin`, and was never assigned — so both were always recomputed (dead, but safe), and simply assigning it would have made a value requested after a gradient at the same point be answered with the gradient. `dimPrevious` was never initialized at all.

As for the one-dimensional types, each quantity now has its own key, and `interpolationReset` — called from `create`, `extend`, `populate` and `destroy`, and after each direct write to `zv` in `Table_Caches` — invalidates them together.

## Deliberately not done

Per-axis scaling composition (step (c) of the issue): it should be driven by an actual call site wanting a mixed scaling rather than built speculatively. #115 is therefore referenced rather than closed.

## Testing

`source/tests/tables.F90` gains `Test_Table_2D`, which runs one set of assertions over **both** concrete types through a `class(table2D)` — impossible before this change, and itself the demonstration that (a) worked. It covers accessor agreement, interpolation at tabulated points, gradients against a finite difference of the interpolant, and the memoization interleavings above. `source/tests/table_caches.F90` gains a `table2DLinLinLin` cache round trip and a cross-type rejection.

| check | result |
| --- | --- |
| `tests.tables.exe` | 70/70 |
| `tests.table_caches.exe` | 21/21 |
| `tests.spherical_collapse.nonlinear` / `.determinism` | 1/1, 30/30 |
| `tests.linear_growth.baryons.EdS` / `.cosmological_constant` | 10/10, 8/8 |
| `tests.halo_mass_function.environment_averaged` | 1/1 |
| `Galacticus.exe` + `quickTest.xml` | exit 0, HDF5 written |
| percolation model, 4 threads | SUCCESS |
| Python test suites | 1026 passed, 1 skipped |
| whole-tree source formatter round trip | passes |

One note: `xs()` and `x(i)` differ by ~1 ulp for logarithmic tables, because at `-O3` gfortran routes the array `exp` through a vectorized libm path and the scalar through another. That is pre-existing and shared with `table1DLogarithmicLinear`, so the abscissa assertions use `relTol=1.0d-14`; the values, which are copied rather than transformed, are still compared bit-exactly.

Refs #115.

https://claude.ai/code/session_01APyicKQo1FqGWCsgkMP8jk


abensonca:
## CI failure in `Validate-Baryonic-Suppression` — diagnosed and fixed in c0fad6248a

The failure was caused by this branch, specifically by the memoization change in 1d433a15d6.

**Mechanism.** The memoized value and gradient of `table2DLogLogLin` were given a table (and dimension) key of their own, but both guards still rested on the interpolation *factors* which the two quantities share. Evaluating either at a point advances those factors, so the other — memoized at an earlier and different point — was answered at the new one:

```
interpolate        (x1,y1,  1)   zPrevious = Z(x1,y1,1)
interpolateGradient(x2,y1,1,1)   advances xLinearPrevious to x2
interpolate        (x2,y1,  1)   guard hits, returns Z(x1,y1,1)   ← wrong point
```

`linearGrowthBaryonsDarkMatter` interleaves exactly these calls on its D(k,t) table, so growth factors came back from the wrong point. That corrupted σ(M), and thence the halo concentration, until a root find drove the NFW profile to the floating point exception seen in the log.

The one-dimensional types are keyed the same way and are safe from this only because they hold no shared factors at all — `Table_Linear_1D_Interpolate` finds its bracketing index in a *local* variable. The previous commit copied their key-splitting without that property.

**Fix.** Each quantity is now keyed on the point at which it was computed, entirely independently of the shared factors.

**Why the tests missed it.** They interleaved value and gradient only at a single point, where a key resting on the shared factors cannot diverge. They now interleave across differing points, in both directions. Both new assertions were written first and confirmed to fail against 1d433a15d6, returning precisely the stale values predicted.

**Verification**

| check | result |
| --- | --- |
| `tests.tables.exe` | 74/74 |
| `tests.table_caches.exe` | 21/21 |
| `tests.spherical_collapse.nonlinear` / `.determinism` | 1/1, 30/30 |
| `tests.linear_growth.baryons.EdS` / `.cosmological_constant` | 10/10, 8/8 |
| `tests.halo_mass_function.environment_averaged` | 1/1 |
| full `validate-baryonicSuppression.py`, 4 threads | all three models complete, every χ² target met |

The validation run that previously died now reports, among others, `withBaryons` at z=9.27 χ²=7.365 < 10.000 and at z=0.00 χ²=1.949 < 3.000.

https://claude.ai/code/session_01APyicKQo1FqGWCsgkMP8jk

PR #1483, opened 2026-09-13
ci: bound and prune superseded bleeding-edge release   assets

abensonca:
Follow-up to #1475.

`tools.tar.bz2` and `toolsMacOSM1.zip` were still listed among the assets mirrored from `bleeding-edge` onto each version tag, but `Deploy` stopped publishing either when the tools moved to a zstd tar — the macOS zip now exists only as the format `xcrun notarytool submit` accepts, and the Linux bz2 not at all.

Because `gh release upload --clobber` never deletes, both sat frozen on the rolling release at whatever commit last produced them:

| asset | size | last refreshed |
|---|---|---|
| `tools.tar.zst` | 1.83 GB | today |
| `toolsMacOSM1.tar.zst` | 1.78 GB | today |
| `tools.tar.bz2` | 1.74 GB | **2026-08-19** |
| `toolsMacOSM1.zip` | 2.06 GB | **2026-08-19** (now deleted) |

Mirroring such a copy onto a new tag would have published a 1.7–2 GB archive built from an unrelated commit alongside the correct one, and recorded it in that release's `SHA256SUMS`. This is the same trap the macOS Intel entries had, removed in abc0b0bf8d.

## What this changes

- Drop `tools.tar.bz2` and `toolsMacOSM1.zip` from the mirror list.
- Rewrite the comment above the list to state the invariant — only names `bleeding-edge` still refreshes belong in it — so a superseded name gets dropped rather than kept "just in case". The previous comment said the opposite, and that is how these two survived.

## What this does not change

Releases cut while either archive was still published keep their own copies and stay installable. The launcher falls back to a legacy archive name (`download._tools_archive`, driven by the `tools_legacy` entries in `platforms.py` — both deliberately retained) and reads each release's own assets, not this list. `cicd.yml` also still *builds* `toolsMacOSM1.zip` as the notarization submission format; it simply is not published to the release.

## Housekeeping done alongside

The retired macOS Intel assets and the stale `toolsMacOSM1.zip` have been deleted from `bleeding-edge`, freeing ~6 GB:

```
Galacticus_MacOS.exe  Galacticus_MacOS.zip  toolsMacOS.tar.zst
toolsMacOS.zip        debugSymbolsMacOS.zip toolsMacOSM1.zip
```

The superseded `tools.tar.bz2` asset is left in place for now — say the word if it should go too.

## Verification

`publish.yml` parses as YAML; launcher test suite passes (131 tests), confirming the legacy-name fallback is untouched. The mirror step itself runs only on a version tag push, so its first real exercise is the next release.

https://claude.ai/code/session_017NbMgaUFY2MztyBkWkN2fe


abensonca:
## Scope update: this PR now carries a second, independent commit

The description above covers only the first commit. Adding the second here rather than rewriting the description.

### 2. `ci(notarization)`: bound the per-commit bookkeeping assets (11b8361)

Notarization tracks each `master` deploy with two small JSON assets keyed by commit — `notarization-requests-<sha>.json` (the `notarytool` submission ids) and `notarization-complete-<sha>.json` (written once the result reached Slack, so a later poll does not report it twice). Same `--clobber` behavior as the archives in commit 1, same result: one pair per deploy, indefinitely. There were **37 commits' worth, back to April**, making up most of the release's asset list.

Only the pair for the current `master` HEAD is ever read — `notarizationRetrieve.py` resolves both names from `git rev-parse HEAD` — so every older pair is history with no reader.

New `scripts/aux/notarizationPrune.py`, run by the `Notarize-MacOS` poller after it has reported and marked the current result. It keeps the HEAD pair, the `--keep` (10) most recently updated commits, and anything newer than `--min-age-days` (7); the rest are deleted in pairs.

The two guards each cover a distinct way the other could be defeated:

- **HEAD explicitly**, because it may also be the *oldest* pair present (a long-lived `master` with no new deploy). Losing its `requests` file would reduce the next poll to "no-request" and the result would never be reported — silently, since that path exits 0 by design.
- **The age floor independently of the count**, because the count is per commit rather than per day: a burst of deploys could otherwise push a pair that is only hours old out of a count-based window.

Deletion is by pair, not by asset: removing a `complete` marker while its `requests` sibling remained would, for HEAD, invite the duplicate Slack notification the marker exists to prevent.

The step is `continue-on-error` — housekeeping must not turn a successful notarization poll red; the script annotates what it could not delete and the next scheduled run retries. Assets are read from the paged `releases/{id}/assets` endpoint rather than the release object, which carries only the first page — a limit this release had reached because of these very assets.

#### Verification

- Nine tests in `scripts/aux/tests/test_notarization_prune.py` pin the policy (HEAD protection, the age floor against a deploy burst, oldest-first ordering, a half-pair from a deploy whose notarization never completed, an empty release, a negative `--keep` not inverting the policy) and the grouping (only the `<sha>`-keyed pair is ever grouped, so the sweep can never see a binary; a pair is as old as its newest member).
- Full `pytest` suite: 1145 passed, 1 skipped.
- `--dry-run` against the live release selects 27 superseded commits (42 assets) and keeps 10 — newest pruned pair 30 days old against a 7 day floor, today's HEAD retained:

```
total commits: 37 | prune: 27 | keep: 10

KEPT (newest first):
  67a5e89a0dfd  2026-09-13  age   0d
  0166a1428931  2026-09-01  age  12d
  d6a011bdd47d  2026-08-31  age  13d
  b97d472f50b1  2026-08-25  age  19d
  0ab8e753dde8  2026-08-24  age  20d
  5e69da91e1df  2026-08-22  age  22d
  0f386049a46f  2026-08-20  age  24d
  544daf1fc5ea  2026-08-19  age  25d
  9b6b0b129609  2026-08-15  age  29d
  0f36c1956a76  2026-08-14  age  30d

newest pruned: 30d old  (floor is 7d)
```

The prune takes effect on the poller's first scheduled run after merge; nothing has been pruned by it yet.

### Manual deletions already done

`tools.tar.bz2` has now also been deleted, bringing the total freed on `bleeding-edge` to ~7.7 GB. All were verified stale beforehand — the Intel assets untouched since 2026-09-01 and absent from the regenerated `SHA256SUMS`; the two superseded archives frozen at 2026-08-19 while their zstd replacements refreshed today:

```
Galacticus_MacOS.exe   Galacticus_MacOS.zip   toolsMacOS.tar.zst
toolsMacOS.zip         debugSymbolsMacOS.zip  toolsMacOSM1.zip
tools.tar.bz2
```

https://claude.ai/code/session_017NbMgaUFY2MztyBkWkN2fe

Issue #893, opened 2025-08-04
Improve/Unify dust attenuation modelling


abensonca:
Currently, dust attenuation is implemented by the `stellarSpectraDustAttenuation` class - but this is quite simple. It allows for wavelength and age dependence, and is normalized by passing a `vBandAttenuation`.

* This class is used for more than just stellar spectra, so is poorly named;
* It is quite simplistic in what it assumes can affect dust attenuation;
* The`nodePropertyExtractorLmnstyStllrCF2000` class (re-)implements its own version of the Charlot & Fall (2000) dust attenuation model - not good to have duplicated implementation - remove dust attenuation from this and have it just return the unattenuated luminosity.

Create a new class which provides a much more versatile dust attenuation framework. Then, create a new `nodePropertyExtractor` which can be applied to any extracted luminosity (broadband, emission line, SED) and simply applies dust attenuation to it.

 * New `dustAttenuation` class:
   * Re-implement the current set of `stellarSpectraDustAttenuation` models:
       * Create an abstract subclass of `dustAttenuation` which simply computes the `vBandAttenuation` and has an abstract method that applies the relevant extinction curve, then implement the current models as subclasses of this abstract class.
   * Requirements:
       * Computes attenuations directly from `node` properties (so must receive `node` as an argument)
       * Can be passed different components (`disk, spheroid, AGN`) and work with them.
       * Can apply to broad-band luminosities (e.g. total and recent).
       * Can apply to time-resolved stellar populations.
       * Can apply to emission lines.
       * Allow for a specific inclination angle or averaged over all angles.
* New `nodePropertyExtractorDustAttenuation` class
    * Apply attenuation to any other `nodePropertyExtractor` that provides a luminosity, output it with a suitable suffix (e.g. `dustAttenuated` - perhaps with some extra descriptive suffix provided by the used `dustAttenuation` class).
        * Optionally still output the unattenuated luminosity also.
    * This would need to handle scalars, tuples, arrays, etc., so probably needs to be a subclass of the `nodePropertyExtractorMulti` class.
        * This is useful anyway - can wrap a bunch of luminosities in single dust attenuator.
    * The wrapped `nodePropertyExtractor` would need to provide methods that give:
        * component information: `disk/spheroid/blackHole/NSC/total` (note that a given attenuator could refuse to apply to, e.g. `total`)
        * viewing angle information?
        * a data structure that details the luminosity produced split by:
            * (rest-frame) wavelength
            * component?
[...]

PR #1485, opened 2026-09-14
feat(dust)!: add a dustProperties class shared by dust attenuation models

abensonca:
## Summary

Phase 4a of #893 (dust emission): a `dustProperties` class that supplies the dust-to-metals ratio, the V-band extinction opacity and the dust mass to every dust model, so the attenuation and the forthcoming dust emission models use the same dust.

Previously each attenuator set its own normalization: a hard-coded Savage & Mathis constant in `screenSurfaceDensityMetals`, a bare coefficient in `birthCloud`, and a separate `dustToMetalsRatio` in `atlasCompendium`. A greybody or template emission model needs the dust mass. If that mass came from somewhere else, changing a screen's normalization would change the reddening but not the infrared, and the IR spectrum would be silently wrong.

### Changes

- **New `dustProperties` functionClass** (`source/dust/properties/`) with methods `dustToMetalsRatio(node,component)`, `opacityExtinctionV()` (cm² g⁻¹) and `massDust(node,component)`. Its implementation `dustPropertiesSimple` defaults to a ratio of 0.44 (Popping et al. 2017) and the opacity for which that ratio reproduces the Savage & Mathis (1979) calibration. **With the defaults, results are unchanged.** `componentGasProperties` moves into this module.
- **`screenSurfaceDensityMetals`:** τ_V = C κ_V f_dust Σ_Z. `coefficient` is kept, and is now documented as a geometric factor that doesn't change the dust mass.
- **`birthCloud`:** `coefficient` is replaced by `densitySurfaceGas`, the gas column through a cloud. Its default, 22.840 M☉ pc⁻², gives τ_V = 1 at local ISM metallicity with the default dust properties.
- **`charlotFall2000`:** keeps its dimensionless coefficients, and passes one dust properties object to both components.
- **`atlasCompendium`:** takes its dust-to-metals ratio from `dustProperties`. It **errors** if that object's opacity differs from the tabulation's by more than 10⁻⁶ relative, and the documentation shows a nested `dustProperties` block carrying the file's opacity.
- **Analyses that build `charlotFall2000` internally** now take a `dustProperties` object: `outputAnalysisColorDistributionSDSS`, `outputAnalysisLuminosityFunction` and its subclasses `MonteroDorta2009SDSS` and `StefanonMarchesini2013`.
- **Migration:** a `dust_properties` function in `parametersMigrate.py`.
  - A `birthCloud` `coefficient` becomes `densitySurfaceGas`, multiplied by 22.840.
  - An `atlasCompendium` `dustToMetalsRatio` moves into a nested `dustProperties`. The opacity can't be filled in automatically; the model reports the required value.

### Breaking change

`birthCloud`'s `coefficient` and `atlasCompendium`'s `dustToMetalsRatio` are removed. Parameter files migrate automatically, except for the compendium opacity noted above. No parameter file in `parameters/` uses either.

## Test plan

- [x] `test-dust-attenuation-parity.py`: all SUCCESS, including three new pairs that must agree to round-off. All three agree to 0.0:
  - screen `coefficient` ×2 vs dust-to-metals ratio ×2;
  - birth-cloud column ×2 vs ratio ×2;
  - `charlotFall2000` vs the explicit `sequence` it stands for.
- [x] `test-dust-analyses-SDSS.py` and `test-luminosity-function-Halpha.py` pass.
- [x] `test-parameter-migration.py` has a new dust case, and passes with the `migrations.xml` entry.
- [x] Python unit tests: 1150 passed, 1 skipped.
- [x] **No change in results:** the SDSS, Hα and parity models were run single-threaded with an executable built before this change and with this branch (a repeated run is bit-identical).

  | Model | Largest relative difference |
  |---|---|
  | SDSS analyses | 0 |
  | Hα luminosity function | 1.5×10⁻¹⁵ |
  | Parity model (78 datasets) | 2.7×10⁻¹⁶ |
- [x] `auditTestMarkers.py --check` and `extractContributors.py --check` are clean.
- [ ] Compendium opacity check: not exercised in CI (the tabulations download from Zenodo).

https://claude.ai/code/session_01GRjkvvhtWPU5mGK5RWWKv9


abensonca:
Merge note: please merge this with a merge commit rather than squashing. `scripts/aux/migrations.xml` keys the `dust_properties` migration to commit `93b7a61e69`; a squash merge would give that change a new hash and the migration would never run. (Phase 4b is being developed on a branch stacked on this one.)

PR #1499, opened 2026-09-15
Physics derivation checks: three bug fixes, two physics corrections, and independent reference tests

abensonca:
Nine commits from an exercise in re-deriving cited physics from the source papers, working
through `derivation-check-candidates.md`. Three bugs, two physics corrections, one
performance change, and three unit tests which compare Galacticus against independent
references rather than against itself.

## Two changes reviewers should look at first

**1. Blitz & Rosolowsky (2006) characteristic pressure — this changes both reference
models.** `[pressureCharacteristic]` defaulted to `4.54` and was used as a linear value of
P₀/k in K cm⁻³. Table 2 of the paper gives P₀/k in units of 10⁴ K cm⁻³ (the "Mean" row
being 3.5, with α = 0.92, their eqn. 12), and log₁₀(3.5×10⁴) = 4.544 — so the default was
log₁₀(P₀/k) used as though it were linear, making the effective P₀ about **7,700× too
small**. Disks were consequently molecular at essentially all radii: in a Milky Way-like
disk f_H₂ stayed at 1.00 out to six scale lengths (Σ_gas = 0.4 M⊙ pc⁻²), where the paper's
value gives 0.34 at three and 0.004 at six. The molecular fraction was also `min(R,1)`
rather than the paper's `R/(1+R)` (their eqn. 21).

Reference models will need recalibrating. The Milky Way validation model moves as you would
expect when a calibrated model loses a large error in one of its inputs — structural
relations improve, abundances degrade (table below).

**2. The spherical scaler mass inversion — silently wrong for every dark-matter-only
halo.** `massDistributionSphericalScaler%radiusEnclosingMass` multiplied by
`factorScalingMass` where it must divide, so it returned the radius enclosing f_DM²m ≈
0.71m instead of m. Affected callers include Gnedin (1999) tidal heating and Petts (2015)
dynamical friction (half-mass radii), both Kummer (2018) SIDM classes, orphan tracing,
`radius_definition`, and the bound-mass, Einstein-radius and tidally-truncated-NFW
extractors. Calls passing `massFractional` were unaffected. The cylindrical scaler was
already correct.

## The rest

| Commit | Change |
|---|---|
| `910f879` | Spherical scaler mass inversion (above) |
| `c5cbe88` | `zhao1996MassTotal` was missing a factor of π and used the wrong Γ argument |
| `d76c838` | King (1962) fallback guard compared total bound mass against dark virial mass, so the first ~16% of stripping had no effect; description corrected |
| `85059c1` | Benson (2005) stored velocity constants were 0.14% and 0.07% off re-integration of the coded fit |
| `b9c98aa` | Blitz (2006) pressure and molecular fraction (above), plus removal of the tabulation |
| `61133a6` | Krumholz (2009) applied the hydrogen mass fraction where KMT09 define total gas |
| `6e9d482` | Extended ETHOS `[cW1]`/`[beta1]` defaulted to 0, giving c_W = 0 or ∞ rather than a constant |
| `20435db` | Cardelli (1989) docstring closed a `:math:` role with `$` |
| `b3a6830` | Three new unit tests |

## Why the Blitz tabulation was removed

Its analytic solutions evaluated the disk integral in a "fully molecular" regime (f = 1)
and a "partially molecular" regime (f = R), which together are exactly the superseded
`min(R,1)`. With `R/(1+R)` the normalization of R no longer divides out, so they cannot be
rescaled. A fourth table axis would need 8–10 decades, taking each table from 70–76 MiB to
**2–7 GB**; a universal collapse of the shape fails (31% spread); a moment expansion about
the transition radius converges too slowly (3–4% at one term, since the transition width
0.54 is not small against the weight's scale of ~1). Fixed-order Gauss–Legendre split at
the transition radius reaches 1.8×10⁻⁷ at 24 nodes per subinterval, against the 10⁻³ the
caller actually integrates to, so intervals are now integrated numerically.

Cost, master against this branch: `checkpointingNoCheckpoints` 25.78 ± 2.58 s → 28.68 ±
1.48 s (six runs per side); `benchmark_milkyWay` 18.55 s → 20.75 s; `validate_milkyWay`
589.7 s → 474.4 s. The last is *not* attributable to the integration — that model's
evolution changes under the new physics — so the honest figure is **of order 10%**. Peak
memory fell 1.5 GB on the first model, more than one table, because the tabulation was
held per thread. `Blitz2006.F90` goes from 1465 to 718 lines.

## Verification

- **New tests:** King1962 12/12, ETHOS 8/8, Zhao1996 11/11. The first found bugs 1 and 3;
  the third found bug 2.
- **Existing tests:** `mass_distributions` 163/163, `dark_matter_profiles.Zhao1996` 15/15,
  `satellites.tidal_stripping.radius.soliton_limited` 3/3.
- **Models run clean** (no fatal/aborted/unrecognized-parameter/ODE-failure markers):
  `validate_milkyWay`, `benchmark_milkyWay`, `molecularCooling`, `barInstabilityFPE`,
  `noninstantaneous_recycling`, `adaptiveSFHLengths`,
  `test-star-formation-histories-fixedAge`.
- **Milky Way validation**, Galacticus' own stored `logLikelihood`, smaller is better:

  | Analysis | Before | After | Change |
  |---|---|---|---|
  | Luminosity function | 60.96 | 72.69 | +19.2% |
  | Mass–metallicity | 39.29 | 42.86 | +9.1% |
  | Mass–size | 16.42 | 8.94 | −45.6% |
  | Mass–velocity dispersion | 1.77 | 0.61 | −65.8% |
  | Occupation fraction | 22.55 | 22.20 | −1.6% |
  | Stellar mass function | 49.13 | 60.15 | +22.4% |
  | Stellar mass–halo mass | 20.00 | 13.09 | −34.5% |
  | **Total** | **210.12** | **220.52** | **+5.0%** |

The reference scripts backing these tests are in `galacticusorg/galacticusDevTools`
(`referenceKing1962.py`, `benson2005Check.py`, `ccm89Check.py`, `sfImpact.py`).

## Follow-ups, not in this PR

- **`[useTabulation]` needs a migration entry.** It is removed from the code and from the
  12 parameter files which set it, but `scripts/aux/migrations.xml` has no entry yet; it
  must reference `b9c98aa`, the commit which removed it.
- The analytic-interval machinery in `integrated_surface_density.F90` is now unused by
  every class, since Krumholz always reports intervals as non-analytic.
- Reference models need recalibrating after the Blitz change.
- Noted but not fixed: Li (2020) lacks an `abs()` on the radial velocity in its
  distribution function (latent — correct for its only caller); Benson (2005) uses only the
  z = 0 fit without saying so; Jiang (2014) carries the paper's negative γ in one bin;
  `ETHOSValue` returns 0 at k ≤ 0; two ETHOS docstring/guard inconsistencies; King1962's
  `massEnclosedBySphere(0) >= 0` guard is dead.

https://claude.ai/code/session_01RqTUNnEGzs1jgSE4a9Z6ZE


abensonca:
# :warning: The bug fixes in this PR plausibly affect Milky Way and dark matter-only subhalo validation model physics :warning:

The results of these validation models should be checked carefully once CI is finished for this PR. Running the Milky Way validation model locally suggests that the changes are not huge (most significant for the lowest mass galaxies, particularly sizes, velocity dispersions, and metallicities). Plots below show the changes.

localGroupLuminosityFunction
localGroupMassMetallicityRelation
localGroupMassSizeRelation
localGroupMassVelocityDispersionRelation
localGroupOccupationFraction
localGroupStellarMassFunction
localGroupStellarMassHaloMassRelation

PR #1511, opened 2026-09-16
fix(satellites): correct two mass-normalization errors in satellite evolution, and verify the rates and their assembly

abensonca:
> **This PR changes physics, not just tests. Please check model results before merging.**
>
> It corrects two errors in the rates governing satellite orbital decay and tidal stripping. Both change satellite evolution wherever satellites are evolved, so model outputs will move. A green CI run is necessary but **not sufficient**: where a validation target shifts, someone needs to judge whether the new value is right, not simply update it. See [What a reviewer needs to check](#what-a-reviewer-needs-to-check) below.

Phase 3, item 9 of the physics verification plan — completing it. Stacked on #1507, which is stacked on #1505, #1500 and #1496.

## The two fixes

A satellite's bound mass is a **total** mass — `satelliteMassBoundInitializorBasicMass` sets it to the node's basic mass. But `node%massDistribution()` is normalized to only the dark matter fraction of that mass. Measured for a 10¹⁰ M⊙ satellite, mass within its virial radius:

| accessor | mass |
| --- | --- |
| `node%massDistribution()` | 8.4364e9 |
| `(componentTypeDarkHalo, massTypeDark)` | 8.4364e9 |
| `(componentTypeDarkMatterOnly, massTypeDark)` | **1.0000e10** |
| `satellite%boundMass()` | 1.0000e10 |

Any expression combining a bound mass with a profile mass must therefore use the dark-matter-only profile. Two did not:

- **`satelliteTidalStrippingZentner2005`** computed the mass outside the tidal radius as `boundMass - massEnclosedBySphere(radiusTidal)`, with the enclosed mass from the baryon-corrected profile. A satellite whose tidal radius reached its own virial radius still had about **16% of its bound mass "outside" that radius**, and so lost mass although nothing was stripping it.
- **the extended-mass suppression factor of the Chandrasekhar integral**, `min(1, M(<r)/M)`, divided a dark matter mass by a total bound mass. It **saturated at 0.844 rather than unity**, suppressing the dynamical friction acceleration of every satellite by the baryon fraction however compact it was.

Both now take `node%massDistribution(componentTypeDarkMatterOnly, massTypeDark)`.

#1499 corrects the same class of mixing in a neighbouring branch of the King (1962) tidal radius, which is what suggested looking here.

## What a reviewer needs to check

The fixes are verified against independent references. **What is not verified is their effect on models.** These classes are used by many bundled parameter files, including the tidal track validations, the checkpointing tests, the constrained merger trees, the decaying dark matter models and lightcone crossing. No production model was run locally.

Both fixes push in a definite direction, which is worth holding in mind when reading any shifted result:

- satellites now experience **more** dynamical friction (the suppression factor no longer caps at 0.844), so they should sink faster;
- satellites that are barely stripped now lose **less** mass, or none, where before they always lost some.

So subhalo mass functions, merger timings and anything downstream of them may move. Where a validation target changes, the question is whether the new value is *better*, not merely different — the previous values were produced by the erroneous normalization. If something moves in the *opposite* direction to the above, that is worth investigating rather than accepting.

## The verification — two tests, neither subsuming the other

**`tests.satellite_orbit_rates.exe`** compares the rates **pointwise**: 14 phase-space configurations, 56 assertions, against `satelliteOrbitRates.py`.

**`testSuite/test-satellite-orbit-evolution.py`** integrates one satellite's orbit and mass loss in a static host potential and compares the **trajectory**, against `satelliteOrbitEvolution.py`.

Both are needed. A pointwise comparison cannot catch a rate which is correct but enters the equations of motion with the wrong sign, factor or units; an integrated trajectory alone cannot distinguish such a mistake from a wrong rate. The second reference imports its rate functions from the first, so the two cannot drift apart.

| quantity | agreement | tolerance | basis for the tolerance |
| --- | --- | --- | --- |
| friction acceleration | 6.7e-5 | 5e-4 | does not use the tidal root finder; floor is the Jeans quadrature |
| tidal radius | 2.1e-4 | 2e-3 | `radiusEnclosingDensityNumerical` has a relative tolerance of 1e-3; doubled, per #1499 |
| tidal mass loss rate | 1.6e-3 | 5e-3 | inherits the radius error amplified by up to ~8.5× |
| orbital radius, integrated | 1.9e-4 | 1e-3 | the two integrators, not the physics |
| bound mass, integrated | 1.2e-4 | 1e-3 | as above |

The mass-loss amplification is itself a consequence of the fix: while `massOuter` carried its spurious floor of the baryon fraction it was never small, and so was never numerically delicate. Correcting the normalization made it a genuine difference of comparable masses. A fix can make a quantity harder to verify.

## Controls

Six, all decisive, each with a distinct signature.

Rates: Ω_b 0.0493 → 0.0400 fails 45 of 56; a 5% Coulomb logarithm change fails exactly the 25 non-zero acceleration assertions while all 28 tidal ones pass; a stripping efficiency change fails exactly the 14 mass-loss assertions.

Trajectory: removing `cosmicTime` freezes the satellite and fails all four assertions; removing dynamical friction fails both value comparisons while both guards correctly pass; removing tidal mass loss leaves the bound mass exactly unchanged.

The trajectory test also carries **two guards against a degenerate pass**. Were the satellite frozen at its initial conditions, every value comparison would still be made — and against a reference generated from those same initial conditions it could even pass. It therefore requires that the orbit is resolved (apocenter more than twice pericenter) and that mass is actually lost. This is not hypothetical: a frozen satellite is exactly what happens when `cosmicTime` is absent.

## Related

Issue #1512 records a silent-failure mode found while building the orbit model: a malformed rank-1 property in a fully-specified tree drops the node, and the run does nothing and exits 0.

## One question left open

King (1962) was deliberately **not** changed. It is the derivation session's item with #1499 in flight, and it asks a different question — where the satellite's *physical* mean density falls to the tidal density — so the baryon-corrected profile is arguably right there. That does leave an asymmetry: the tidal radius comes from one profile while `Zentner2005` reads the enclosed mass at that radius from the other. Flagged in the script and the test rather than resolved unilaterally.


abensonca:
# :warning: Satellite tidal physics and dynamical friction affected - be sure to review validation results before merging :warning:

abensonca:
Milky Way model results do not seem to be significantly affected by these changes. And dark matter only models are not affected at all since $\Omega_\mathrm{b}=0$ in those.

PR #1519, opened 2026-09-17
Check four spherical mass distributions against independent integration of their own densities

abensonca:
The first piece of Tier 3 of the physics derivation-check programme: re-deriving cited physics from
the source papers, rather than testing Galacticus against itself. Branched off `master`.

Four spherical mass distributions are defined by their density alone and evaluate the enclosed mass,
potential and everything else from closed forms. Nothing checked those closed forms against anything
independent. `tests.dark_matter_profiles.generic` compares each profile's analytic results against
*its own* numerical integrals — 16 assertions per profile — which establishes that the closed forms
match the density as coded, but not that either is right. The two `solitonNFW` assertions in
`tests.mass_distributions` only check that `densityGradientRadial` agrees with itself with and
without its `logarithmic` argument.

Two new tests, 31 assertions, compare all four against an independent reference which integrates
each density numerically (`massDistributionProfilesCheck.py`, added to
[galacticusDevTools](https://github.com/galacticusorg/galacticusDevTools)), and assert the NFW
limits exactly.

### Einasto and Burkert

**Einasto is clean** — density, enclosed mass, potential and total mass all reproduce independent
integration to ~10⁻¹³, for shape parameters α = 0.12, 0.17 and 0.30.

**Burkert had three defects**, all in or about `massEnclosedScaleFree`:

1. Its small-radius series was (4π/3)R³(1−R). Since ρ = 1−R+O(R⁴), the mass is 4π(R³/3−R⁴/4), so
   the correction term was 4/3 too large.
2. More consequentially, that series engaged only below 10⁻⁴r_s, but the exact solution is *already*
   wrong by 6.2×10⁻⁵ there: its terms are of order R while their sum is of order R³, and
   `log(1.0d0+radius)` rounds its argument to unity plus an ulp first. The band from 10⁻⁴ to 10⁻²r_s
   was served accurately by neither branch — outside where one would look for a small-radius problem.
3. The documented potential was the exact negative of the implemented one. The code is right: it
   gives Φ(0) = −π², which follows from ∫₀^∞ sρ ds = π/4.

Fixed by extending the series — since (1−R)(1+R+R²+R³) = 1−R⁴, the density is (1−R)(1+R⁴+R⁸+…) and
the mass is 4π(R³/3−R⁴/4+R⁷/7−R⁸/8+…) — and raising the switch radius to 10⁻². Error below the scale
radius falls from up to 6.2×10⁻⁵ to ≤2×10⁻¹⁶.

**No model output changes**: nothing under `parameters/` or `testSuite/` selects the Burkert profile.

### cusp-NFW and soliton-NFW

The limits are asserted *exactly* rather than approximately. The cusp-NFW density of Delos & White
(2025) is √(1+y²r_s/r) times the NFW density of the same ρ_s and r_s at **every** radius, reducing to
equality at y=0 — sharper than checking agreement at large radius. Beyond `radiusSoliton` the
soliton-NFW density must equal NFW and the mass it adds must equal the mass NFW adds. All to 10⁻¹².

**One defect found.** `cuspNFWMassEnclosedBySphere` selects between four solutions, one of which
approximates ρ by dropping its (1+r/r_s)⁻² factor, so its error grows in proportion to radius —
1.2×10⁻³ at the top of the range over which it was used, where the full solution is accurate to
10⁻⁹–10⁻¹². It is genuinely needed, but only far below that, where the alternatives lose accuracy to
cancellation. It now has its own threshold of 10⁻⁵, chosen by scanning: that minimizes the largest
error over the whole range of cusp amplitude and radius, 6.7×10⁻⁴ → 2.6×10⁻⁵, and below 10⁻⁸ for the
amplitudes y ≳ 0.05 which occur in practice.

A comment records that the branches must **not** be reordered to let the NFW-limit case take
precedence: that solution evaluates ln(1+x)−x/(1+x), which cancels to nothing at small x, and taking
it first makes y=0 wrong by a factor of 165 at r/r_s = 10⁻⁹. Dodging that cancellation is exactly
what the simplified solution is for.

The soliton core is clean (closed form ~10⁻¹², its own series ≤3×10⁻⁸ where used).

### Verification

Each fix has a demonstrated negative control. Both tests fail on the unfixed code, at the predicted
radii and by the predicted magnitudes — for Burkert at four radii by 2.5×10⁻⁷, 2.5×10⁻⁶, 6.2×10⁻⁵
and 2.3×10⁻⁷, which is what identified its two causes as distinct; for cusp-NFW at r/r_s = 10⁻⁴ by
1.2×10⁻⁴ for both cusp amplitudes. The Einasto half was separately mutated (density exponent
2.0 → 2.02, a shape error invisible at r₋₂): the density assertions fail at every radius except the
scale radius, while the mass and potential assertions — from separate closed forms — all pass, which
is the blind spot those density assertions were added to close.

All existing tests which touch these four profiles still pass: `tests.dark_matter_profiles` (100),
`tests.dark_matter_profiles.generic` (82), `tests.dark_matter_halo_radius_enclosing_mass` (6),
`tests.prompt_cusps` (16), `tests.mass_distributions` (163) and
`tests.mass_distributions.tabulated` (27). Unlike Burkert, cusp-NFW *is* selected by model parameter
files, so `test-promptCuspNFW.py` and `test-prompt-cusps-unformed.py` are the ones to watch in CI
here; a full local model run is underway in parallel.


abensonca:
> RST spell-check — misspellings (click to expand)
> 
> Correct the spelling, mark a code identifier up as `likeThis`, or — for a genuine technical term or proper name — add it to `aux/words.dict`.
> 
> ```
> docs/tutorials/models/01-dust-absorption-and-emission.ipynb:465: (µm)  , and the emission of each phase. Warm dust in birth clouds peaks near 60 µm. In the diffuse medium, PAHs produce the emission bands between 3 and 20 µm, and larger grains the far-infrared peak.
> ```

Fixed elsewhere - ignore.

PR #1520, opened 2026-09-17
fix(nodeOperators): make the satellite-merging callback pointer actually thread-private

abensonca:
## Summary
- Two node operators declared a module-scope callback pointer `self_` and then wrote `!$omp threadprivate(self)` — naming `self`, not `self_`. Neither module has a module-scope `implicit none`, so implicit typing quietly invents a real scalar `self`, makes *that* thread-private, and leaves the actual `self_` pointer **shared between OpenMP threads**. Fixed in `satellite_merging/radius_trigger.F90` and `satellite_merging/soliton.F90`, matching the 18 correct `threadprivate(self_)` sites elsewhere in the tree (these two were the only exceptions).
- It compiles silently, with no warning even at `-Wall -Wextra`, which is why it has gone unnoticed.

Measured rather than inferred:

| Check | Result |
|---|---|
| Minimal reproducer, `gfortran -fopenmp -Wall -Wextra` | compiles, **zero warnings** |
| Reproducer symbols | `TLS __m_MOD_self` (the phantom), `OBJECT __m_MOD_self_` (the real pointer, shared) |
| Symbols in a **built `Galacticus.exe`** | `…mergingradiustrigger__MOD_self_` and `…mergingsoliton__MOD_self_` are `OBJECT` (shared); their phantom `self` is `TLS` |
| Controls in the same binary | `…hydrogennetwork__MOD_self_`, `…sersic__MOD_self_`, `…zhao1996__MOD_self_`, `…mergerprogenitorpropertiescole2000__MOD_self_` — all `TLS` |
| After this change | `…mergingradiustrigger__MOD_self_` / `…mergingsoliton__MOD_self_` become `TLS` |

Why the shared pointer is reachable: `differentialEvolution` sets `interrupt = .true.`, `functionInterrupt => mergerTrigger` and `self_ => self`, and the `mergerTrigger` callback dereferences `self_` *later*, after the ODE solver unwinds the interrupt. That path runs under threaded tree evolution — `merger_trees/node_evolver/standard.F90:1055` is called per worker at `merger_trees/evolver/threaded.F90:637` inside the parallel region at `threaded.F90:418`, with forests evolved in parallel at `tasks/evolve_forests/_class.F90:624`. (`parameters/quickTest.xml` reports 2.9 s core-time against 0.866 s wall-time, i.e. trees really are evolved concurrently.)

**Severity: latent, not currently producing wrong results.** No bundled or `testSuite` configuration instantiates either operator twice, so every thread stores the same pointer value and the race is benign as shipped. It would become a non-deterministic wrong-metadata bug for a model that lists the same operator twice — which `nodeOperatorMulti` permits by walking its `processes` list, and which other operators already do (`testSuite/parameters/treeFilterLabels.xml` lists `label` twice). Recording of merged-subhalo Keplerian elements (`recordMergedSubhaloProperties`, `mergedSubhaloIDs`) is what would be attributed to the wrong instance.

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Docs / tooling
- [ ] Other:

## How to test
- `make -j8 Galacticus.exe`, then from `testSuite`: `python3 test-satellite-distance-minimum.py` (its parameter file enables `satelliteMergingRadiusTrigger`; the operator is also used by `parameters/reference/evolutionGalaxyFormation.xml`, `evolutionDarkMatterOnly.xml` and 33 `testSuite` files).
- Confirm the thread-locality directly: `readelf -sW Galacticus.exe | grep -i merging | grep -i self` — both `…__MOD_self_` symbols should now be `TLS`, and the phantom `…__MOD_self` should be gone.

## Checklist
- [x] I ran relevant tests (or explain why not): builds with no new warnings — the two `allowSubsampling` warnings a fuller build emits come from `source/numerical/points/convex_hull.F90` in master, which this change does not touch; `test-satellite-distance-minimum.py` passes on 4 threads; and the symbol table confirms both `…__MOD_self_` symbols are now `TLS` with the phantom `…__MOD_self` gone.
- [ ] If this PR is from a fork: I enabled 'Allow edits from maintainers'

https://claude.ai/code/session_011QhoSB1WnvvmzxinZESrmc


abensonca:
> RST spell-check — misspellings (click to expand)

Fixed elsewhere - ignore.

PR #1522, opened 2026-09-17
Verify the Gnedin (1999) tidal heating chain - rate, accumulation and profile response - and fix four defects it exposed

abensonca:
## Reviewer: this changes physics, so model results must be checked

Three physics fixes here, not just a test. CI going green is necessary but not sufficient — where a
validation target moves, the question is whether the new value is *better*, not merely different.

What moves, and what does not:

| Affected | Why |
|---|---|
| **Models with baryons in satellites** | The Gnedin (1999) orbital frequency now includes the satellite's baryonic mass, so such satellites are heated *less*. |
| **The `satelliteTidalField` property extractor, for eccentric orbits** | It is the one caller passing `includeCentrifugalAcceleration=true`, and the centrifugal term was wrong except for circular orbits. |
| **Models pairing `sphericalSymmetry` with tidal heating** | None are bundled, so this is latent rather than active. |
| **Models setting `coefficientSecondOrder1` or `...2` non-zero** | The heated *density* gains the missing coefficient-derivative term: 1.7-3.8% at the coefficients of `parameters/reference/darkMatterHalosTidalHeating.xml`. Enclosed mass, and so Vmax and Rmax, are unchanged. |
| *Unaffected:* dark-matter-only models | With no baryons the total and dark matter distributions coincide, exactly. |
| *Unaffected:* King (1962) tidal radius, simple tidal mass loss, spheroid tidal heating | All pass `includeCentrifugalAcceleration=false` and use only the radial component. |
| *Unaffected:* the Errani & Navarro (2021) tidal-track calibration | Its metrics are Vmax/Rmax, which depend on enclosed mass rather than density, and every `tidalTrackBestFit_*` file sets a1 = a2 = 0. `test-tidalTracks.py` passes 4/4 with fit metrics intact. |

## What this verifies

Plan item 9 deliberately left out Gnedin (1999) tidal heating. This covers the heating **rate**
pointwise, against `galacticusDevTools/satelliteTidalHeatingRate.py`, which derives

    dQ/dt = (epsilon/3) [1 + (omega T_shock)^2]^-gamma g_ij G_ij

from the impulse approximation rather than transcribing the Fortran. The path-integrated tidal
tensor `G_ij` is set directly, so the rate is checked separately from its accumulation along an
orbit — a comparison of accumulated heating alone could not tell a wrong rate from a correct rate
wrongly accumulated. The accumulation is covered by the second test below.

`tests.satellite_tidal_heating_rate.exe`: **112/112 assertions pass**, largest fractional difference
**4.7e-5** against a tolerance of 2e-4.

Sixteen configurations put the satellite off the coordinate axes with path integrals carrying
off-diagonal elements, so the tensor's orientation and the double contraction are both exercised;
they span an adiabatic correction from 0.92 down to 4.5e-5, so its exponent is tested and not merely
its presence; and they cover the half-mass minimum, the virial-frequency fallback, and a negative
contraction clamped to zero. Three give the satellite a Hernquist spheroid.

`test-satellite-tidal-heating-evolution.py` then covers the **accumulation**: one satellite's orbit is
integrated with tidal heating applied, and the accumulated heat input Q compared against
`galacticusDevTools/satelliteTidalHeatingEvolution.py`, which integrates

    dG_ij/dt = g_ij - efficiencyDecay G_ij / T_orb,   dQ/dt = (epsilon/3) A(omega T_shock) g_ij G_ij

alongside the orbit. The model is `satelliteOrbitEvolution.xml` plus the `satelliteTidalHeating` node
operator, with `darkMatterProfile` left as `darkMatterOnly` so Q accumulates but is not fed back into
the satellite's profile — keeping this a test of the accumulation rather than of the profile's
response, which `tests.dark_matter_profiles.heated.exe` covers analytically. **7/7 pass**: radius
3.1e-6, bound mass 1.6e-4, Q 4.9e-4, against tolerances of 1e-4, 1e-3 and 2e-3. Q rises a
thousandfold through the pericentre passage, so the comparison spans a real dynamic range.

Its three controls: removing the operator leaves Q identically zero (100% off, and the accumulation
guard fires) while the orbit and bound mass still agree, confirming the heating does not feed back;
`efficiencyDecay` 1.0 -> 0.0 puts Q 4.8% out **while every guard still passes**, so the value
comparison has teeth the guards do not; `epsilon` +1% puts it 1.05% out.

### A floor that was misattributed

`test-satellite-orbit-evolution.py` previously recorded that its agreement was limited by the two
integrators. It was not. Tightening Galacticus' ODE tolerance a hundredfold, and its satellite
timestep criterion tenfold, leaves the differences unchanged to three figures. The real cause was that
the reference recomputed the satellite's initial position instead of reading the tree's: the two differ
by 4e-5, because the tree carries the virial radius under `matterLambda` at 13.8 Gyr while that model
runs in a static universe. With the reference taking the tree's values verbatim, agreement in radius
improves **1.2e-4 -> 3.1e-6**, and its tolerance is tightened to 1e-4. What remains in the bound mass
is the King (1962) tidal radius, whose root finder works to a relative tolerance of 1e-3.

## The fourth test: the profile's response

`tests.dark_matter_profiles.heated_tidal_NFW.exe` closes the chain - rate, accumulation, and now what the
accumulated heat does to the satellite's profile. It compares the initial radius, enclosed mass and
density of a heated NFW profile against `galacticusDevTools/satelliteHeatedProfile.py`, over 12
configurations asserted twice: with the second-order coefficients non-zero and with them zeroed, so each
term is exercised separately. Shells expand by 1.01x to 3.07x and the second-order term runs from 1.5% to
57% of the first. **72/72 pass at 1.0e-10**, against tolerances of 1e-7/1e-7/1e-6.

The existing `tests.dark_matter_profiles.heated.exe` sets **all three second-order coefficients to zero**
and tests only an isothermal profile, where a closed-form solution exists. Neither the second-order term -
nor the velocity dispersion and density slope it is built from - nor any profile without an analytic
solution was covered. The fourth fix below lived in exactly that gap.

## The four fixes

**1. The orbital frequency ignored baryons** (`93699a1bfa`). `omega` came from the satellite's dark
matter distribution alone, so a satellite's own galaxy did not contribute. The circular velocity is
now taken from the total mass distribution, at the same (dark matter) half-mass radius.

**2. `sphericalSymmetry` built its tensor in a rotating frame** (`3964dc8312`). It evaluated the
tidal tensor with the satellite placed on the x-axis. The host is spherically symmetric but the
tensor is not isotropic, so this returns the tensor of a frame rotating with the satellite — and the
tidal heating node operator integrates that tensor along the orbit. The pericenter branch necessarily
keeps an axis frame, and now builds a consistent one (satellite on x, moving in y), exact there
because the velocity is purely tangential at pericenter.

**3. The centrifugal term used the velocity, not the angular velocity** (`3964dc8312`). Both classes
built it as `(|v|^2 delta_ij - v_i v_j)/r^2`, where the correct tensor is
`|w|^2 delta_ij - w_i w_j` with `w = r x v / r^2`. Confirmed against a numerical Jacobian of the
centrifugal acceleration; the two agree only for a circular orbit. In the spherically-symmetric form
even the radial component was wrong, using the total speed where the tangential one belongs.

**4. The heated density was inconsistent with its own mapping** (`bc0e9fcf35`). The second-order
coefficient f = a0 + a1 s + a2 s^2 depends on radius through the density logarithmic slope s, but
`tidalSpecificEnergyGradient` differentiated only r and the velocity dispersion, not f. The heated density
is built from that gradient while the enclosed mass comes only from the radius mapping, so the class' own
density did not equal (1/4 pi r^2) dM/dr. Replicating the code's gradient formula in Python reproduced its
densities to eight digits, which pinned the cause exactly; with a1 = a2 = 0 the two gradients agree to
1e-11.

The missing term needs the second logarithmic derivative of the density, added as a new
`massDistribution` method, `densitySlopeLogarithmicGradient`. Its base implementation raises an error
rather than guessing, with analytic implementations for the distributions tidal heating actually reaches:
NFW, cusp-NFW, Zhao (1996), soliton-NFW, finite-resolution (both the general wrapper and its separate NFW
specialization) and exponentially-truncated. Most were found by *running* the bundled models which set a1
non-zero and seeing which aborted. The term, and the call, are skipped entirely when a1 and a2 are both
zero - the default - so distributions without the method behave exactly as before.

Also in that routine: the second-order branch was guarded by `energyPerturbationSecondOrder > 0`, so a
negative f silently dropped the whole term from the gradient. Now `/= 0`.

## Documentation corrections

Three descriptions disagreed with the code, all found by deriving the physics independently rather than by
reading:

- The adiabatic factor was written `(T_shock/T_orb)^2` where the code uses `(omega T_shock)^2` - Gnedin &
  Ostriker's `x = omega tau`, differing by 2 pi inside the square.
- The expansion formula was `1/r_f = 1/r_i - 2 r_i^3 Q/(G M)`; the exponent must be **r_i^2**, which
  dimensions settle independently of the code. The description now leads with the energy condition actually
  solved, notes the closed form holds only without the second-order term, and says that the mapping is
  solved for r_i at a *given* r_f.
- `massDistributionHeatingTidal` gave the second-order term with a coefficient of 2/sqrt(3); the code uses
  sqrt(2). Checked against Benson & Du (2022), the source of the model, which confirms **sqrt(2)**: the code
  was right and the documentation wrong. Corrected, and that paper is now cited (new bibliography entry,
  metadata from Crossref).

In every case the code was right and the documentation wrong.

## Reference script units (`b9c8c6543c`)

Two unit conventions in the existing `galacticusDevTools` reference scripts were wrong, and are fixed
upstream: a Julian gigayear where Galacticus uses a sidereal one (3.8e-5), and the mean matter
density evaluated today rather than at the nodes' epoch of 13.8 Gyr, where a = 0.99997. This
regenerates the values they feed, with **no tolerance changes** — the agreement improves:

| Test | Before | After |
|---|---|---|
| `tests.satellite_orbit_rates.exe` (56/56) | 1.61e-3 | **1.46e-3** |
| `test-satellite-orbit-evolution.py` (5/5), radius | 1.86e-4 | **1.22e-4** |
| `tests.cooling_chain.exe` (8/8) | — | three time-bearing values move by 3.8e-5 |

## Controls

Every fix has a control confirming the test can detect it, each with a distinct signature:

| Control | Result |
|---|---|
| Circular velocity from the dark matter distribution (fix 1 reverted) | exactly the 3 spheroid configurations fail, by 3.9%, 18%, 127% |
| Axis-frame tensor restored (fix 2 reverted) | all 15 off-axis configurations fail; the on-axis one still passes |
| Velocity-based centrifugal tensor (fix 3 reverted) | all 16 fail, for both classes |
| Radial centrifugal from the total speed (fix 3 reverted) | 15 of 16 fail; the exception has a purely tangential velocity, where the forms coincide |
| `epsilon` +1% | all 12 non-zero rates fail, each by 1.0% |
| `gamma` 2.5 -> 2.4 | 12 fail, by up to 49% |
| Heated density with the coefficient derivative omitted | exactly the 7 second-order density assertions fail, by up to 1.8e-3; radius and mass unaffected |
| `satelliteTidalHeating` operator removed (accumulation test) | Q identically zero, guard fires, orbit and bound mass still agree |
| `efficiencyDecay` 1.0 -> 0.0 | Q 4.8% out **while every guard still passes** |

## Tolerance, measured rather than tuned

Every geometric ingredient agrees with the reference to rounding — verified by printing them
individually. What remains is a 7.5e-6 difference in the virial radii (a mean density differing by
2.3e-5), not traced further, which the adiabatic correction amplifies by up to `2 gamma x^2/(1+x^2)`,
approaching 5. That bounds the rate at about 6e-5; measured 4.7e-5; 2e-4 allowed. Any error in the
*form* of the rate changes it by far more, as the controls show.

`parameters/quickTest.xml` runs clean and the build is warning-free.


abensonca:
# :warning:  Review baryoniv validation models for changes before merging - see PR description :warning:

